PR #1, opened 2019-11-13
Adding License

dvenprasad:
📝

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,25 @@
+BSD 2-Clause License
+
+Copyright (c) 2019, Alex's Lemonade Stand Foundation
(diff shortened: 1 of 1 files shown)


cgreene:
We usually use 3-clause. Is there a reason to prefer 2 here?
https://opensource.org/licenses/BSD-3-Clause

jaclyn-taroni:
Most of our repositories use BSD-3 and some of them use dual licensing when they include notebooks, figures, etc. What do you think the contents of this repository will be?

dvenprasad:
Updated it to BSD-3. 

@jaclyn-taroni Contents are going to be code.

PR #27, opened 2020-03-02
WIP Add basic structure for the project

### LINKED ISSUE #26: Initialize Django app
Get a handful of models and their views implemented, determine project layout, and set some conventions for the project.

arielsvn:
close #26 

Bootstrapped the project using https://github.com/agconti/cookiecutter-django-rest then changed the directory structure and removed some features we didn't need.

The project is already setup with Docker and can be executed with `docker-compose up`. The APi restarts automatically every time the code is modified.

There's a `/docs` folder where we can add project documentation that can be deployed with `mkdocs` to github pages. This would be mostly for developers, similar to refinebio's readme.

The project uses django rest framework user authorization, we can evaluate if we can reuse it with OAuth or if we want to roll our own.

I made it so that it's a single app with individual folders for `/models`, `/serializers` and `/views`. Since we were transitioning to something like this in refinebio.

The app contains two models: `User` and `Material` with their corresponding serializers and viewsets.

@davidsmejia @kurtwheeler what do you think? Any comments welcome.

We would still need to add CI configuration.


README.md
```diff
@@ -1,3 +1,25 @@
...
+Start the dev server for local development:
+
+```bash
+docker-compose up
+```
+
+Run a command inside the docker container:
+
+```bash
+docker-compose run --rm web [command]
```
kurtwheeler: What's `rm`? Is there a long-form option for it?

docs/api/users.md
```diff
@@ -0,0 +1,106 @@
...
+```json
+Content-Type application/json
+201 Created
+
+{
+  "id": "6d5f9bae-a31b-4b7b-82c4-3853eda2b011",
+  "username": "richard",
+  "first_name": "Richard",
+  "last_name": "Hendriks",
+  "email": "richard@piedpiper.com",
```
kurtwheeler: :joy:

README.md
arielsvn: Looks like [there isn't](https://docs.docker.com/compose/reference/run/). `--rm` should: "Remove container after run. Ignored in detached mode."

resources_portal/test/factories.py
```diff
@@ -0,0 +1,18 @@
+import factory
+
+
+class UserFactory(factory.django.DjangoModelFactory):
```
kurtwheeler: Wooooaaaahhhhh!!!! I like the looks of this! :+1:

setup.cfg
```diff
@@ -0,0 +1,18 @@
+[flake8]
+ignore = E226,E302,E41,E702,E731
+max-line-length = 110
```
kurtwheeler: 110?

PR #29, opened 2020-03-03
Enable github actions for linting and tests

arielsvn:
## Issue Number

#26 

Testing github actions to run tests

## Types of changes


cgreene:
BTW not sure if you've seen some of the recipes on their marketplace but you can even get the linter to markup things on github if you use that https://github.com/marketplace/actions/lint-action

.github/workflows/push.yml
```diff
@@ -0,0 +1,27 @@
...
+      - uses: pre-commit/action@v1.0.1
+
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build the stack
+        run: docker-compose build
+      - name: Test
+        run: docker-compose run --rm web bash -c "flake8 . && python wait_for_postgres.py && ./manage.py test"
```
cgreene: My impression is that github actions might benefit from having a separate workflow for something like linting and testing. Your linting workflow could be largely derived by the one from the marketplace that I linked. I feel like that's the main value - you can break up large CI processes into small tasks that run on specific triggers. Some might share triggers.

arielsvn:
That looks way better, I'm going to try to install that one instead! 🔥

arielsvn:
Enabled [lint-action](https://github.com/marketplace/actions/lint-action) instead, it's way better! Great recommendation @cgreene 

This should be ready for another review.

Issue #12, opened 2019-11-25
Account Management: Scope, Requirements, Design Decisions 


dvenprasad:
### Context

Users need to manage what kind of notifications they get, how they are contacted etc.
### Problem or idea

![Resources Portal Workflow - Account Management](https://user-images.githubusercontent.com/15315514/69556114-351cf500-0f72-11ea-874b-eefa876d786c.jpg)

Related #6 

### Solution or next step

Whiteboarding session to dig deeper into the details of the workflow and work on layouts for low-fi.

PR #42, opened 2020-03-18
Implement django models

### LINKED ISSUE #37: Implement django models
### Context

We have the database model mostly hashed out. 

### Problem or idea

It's time to implement it within Django.

### Solution or next step

Make all the models and add migration files for them. Bonus would include some serializers/views.

arielsvn:
## Issue Number

close #37 

This adds an initial implementation for the models we discussed. 

I added `TODOs` on the fields where I was unsure of their types, pay special attention to them when reviewing.

This also uses [Django's classes](https://docs.djangoproject.com/en/3.0/topics/auth/default/#permissions-and-authorization) for the Permissions/Roles, we can discuss if we should have our own models instead.

## Types of changes

## Functional tests

Ran migrations locally.

## Checklist

## Screenshots

For reference, I used this screenshot as a guidance

image

And a graphic of the resulting data model is here

image


resources_portal/models/material_share_event.py
```diff
@@ -0,0 +1,40 @@
...
+        on_delete=models.CASCADE,
+        related_name="material_share_events",
+    )
+
+    assigned_to = models.ForeignKey(
+        User,
+        blank=False,
+        null=False,
+        on_delete=models.CASCADE,
+        related_name="material_share_assigned_to_set",
```
davidsmejia: Maybe `material_share_assignments` would be easier to use?

davidsmejia: I mean using `assignments` instead of `assigned_to_set` generally

docs/index.md
```diff
@@ -28,3 +28,13 @@ docker-compose run --rm web ./manage.py createsuperuser
...
+
+# Visualizing the data model
+
+This project includes [django-extensions graph_models command](https://django-extensions.readthedocs.io/en/latest/graph_models.html).
+
+```bash
+docker-compose run --rm web ./manage.py graph_models -a -g > model.dot
+```
+
+At the moment png exports are not setup but these can be generated them from the `.dot` file.
```
kurtwheeler:
```suggestion
At the moment png exports are not setup but these can be generated then from the `.dot` file.
(Either with `dot -Tpng model.dot -o database_diagram.png` or https://convertio.co/dot-png/)
```

kurtwheeler:
What do you think about including that visualization of the database in the documentation somewhere?

resources_portal/models/grant.py
```diff
@@ -0,0 +1,22 @@
...
+        db_table = "grants"
+        get_latest_by = "created_at"
+
+    objects = models.Manager()
+
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    title = models.TextField()
+    founder_id = models.CharField(max_length=80)
```
kurtwheeler:
```suggestion
    funder_id = models.CharField(max_length=80)
```

arielsvn:
> What do you think about including that visualization of the database in the documentation somewhere?

I love that idea, you mean something automatic? it should be possible to have [graph_models](https://django-extensions.readthedocs.io/en/latest/graph_models.html) generate a png that we could update automatically every time the migrations are executed.

However, I couldn't get `pip install pygraphviz` to work in the container. We could address it in a different issue and add a static png now. What do you think? Here I added documentation on how to generate a `.dot` file from which the graph can be generated.

resources_portal/models/material.py
```diff
@@ -10,13 +10,23 @@ class Meta:
         get_latest_by = "created_at"
 
     objects = models.Manager()
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
 
+    material_type = models.CharField(max_length=32)
```
kurtwheeler: Can we use [Django's choices](https://docs.djangoproject.com/en/3.0/ref/models/fields/#choices) here to enumerate the valid values and make sure that we only ever store those. I don't like the way the django choices feature has the "human readable" version and then the version that actually gets stored in the DB, so can you just make them the same, i.e. `]("CELL_LINE", "CELL_LINE"), ...]`

kurtwheeler:
> However, I couldn't get pip install pygraphviz to work in the container. We could address it in a different issue and add a static png now. What do you think? Here I added documentation on how to generate a .dot file from which the graph can be generated.

I think that's a solid plan :+1:

resources_portal/models/material.py
```diff
@@ -10,13 +10,23 @@ class Meta:
...
     url = models.TextField(blank=True)
     pubmed_id = models.CharField(max_length=32, blank=True)
+    url = models.CharField(max_length=255, blank=True, null=True)
```
kurtwheeler: You have this defined twice (also on line 17), and I think I like the other one more. Let's use a TextField that can be blank and can be null.

resources_portal/models/material.py
```diff
@@ -10,13 +10,23 @@ class Meta:
...
 
-    metadata = JSONField(default=dict)
+    additional_metadata = JSONField(default=dict)
 
-    primary_contact = models.ForeignKey(User, blank=False, null=False, on_delete=models.CASCADE)
+    mta = models.CharField(
+        max_length=255, blank=True, null=True, help_text="Contains an url to download the MTA."
+    )
```
kurtwheeler: Can we use a TextField for this one as well?

resources_portal/models/material_request.py
```diff
@@ -0,0 +1,38 @@
...
+    material = models.ForeignKey(
+        Material, blank=False, null=False, on_delete=models.CASCADE, related_name="requests"
+    )
+
+    requester = models.ForeignKey(
+        User, blank=False, null=False, on_delete=models.CASCADE, related_name="material_requests"
+    )
+
+    assigned_to = models.ForeignKey(
+        User, blank=False, null=False, on_delete=models.CASCADE, related_name="assigned_to_set"
```
kurtwheeler: What about `User.inbound_requests` and `User.outbound_requests` for the related names? Is that as clear as it sounds to me?

arielsvn: hmm what would be the difference between `inbound_requests` and `outbound_requests`?

I was thinking of using `assignments` for this one as @davidsmejia suggested

resources_portal/models/material_share_event.py
```diff
@@ -0,0 +1,40 @@
...
+    event_type = models.CharField(max_length=255, blank=True, null=True)
+
+    time = models.DateTimeField()
+
+    user = models.ForeignKey(
+        User,
+        blank=False,
+        null=False,
+        on_delete=models.CASCADE,
+        related_name="material_share_events",
```
kurtwheeler: What if we named this field `created_by` and had its related_names be `material_share_events_created`?

resources_portal/models/material_request.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+
+    mta = models.CharField(max_length=255, blank=True, null=True)
+    irb = models.CharField(max_length=255, blank=True, null=True)
+
+    # TODO: check type of this field
+    executed_mta = models.BooleanField(default=False)
```
kurtwheeler: Hmm, good question. I forget what the purpose of this field was. Do you remember @dvenprasad ?

dvenprasad: The requester will need to sign the MTA once their request for the resource is approved. I think this is to track whether they had signed it or not.

resources_portal/models/organization.py
```diff
@@ -0,0 +1,17 @@
...
+    class Meta:
+        db_table = "organizations"
+        get_latest_by = "created_at"
+
+    objects = models.Manager()
+
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    owner = models.ForeignKey(User, blank=False, null=False, on_delete=models.CASCADE)
```
kurtwheeler:
```suggestion
    owner = models.ForeignKey(User, blank=False, null=False, on_delete=models.CASCADE, related_name="owned_organizations")
```

resources_portal/models/material_request.py
kurtwheeler: Mmm yeah I like that!

resources_portal/models/material_request.py
kurtwheeler: Hmm so does this summarize the flow?
  * An MTA (in the form of a PDF) is uploaded with the material.
  * A requester downloads that PDF, fills it out, and uploads it with their material request.
  * A reviewer approves the request.
  * The requester then reuploads the MTA but with it being signed this time.

If so, should we have the executed_mta field also be a PDF? This way instead of overwriting the original MTA that was uploaded with the request we just have a second one that is signed. One thing that concerns me is that someone could upload a "signed MTA" that isn't actually signed, we won't actually check it, but we'll interpret them uploading it as having signed it. Then when we notify the sharer that the requester uploaded the MTA we can say something like "make sure that the MTA was signed in all required locations" or something like that.

resources_portal/models/user_notification.py
```diff
@@ -0,0 +1,23 @@
+from django.db import models
+
+from .user import User
+
+
+class UserNotification(models.Model):
+    class Meta:
```
kurtwheeler: Can you add a docstring explaining this model? I actually forget exactly how we intended to use it.

resources_portal/models/material_request.py
dvenprasad: No, it is
- Requester makes request, at this point they can download the MTA and read it. 
- A sharer/reviewer approves request
- Requester uploads the signed MTA to the portal (we can send them an email with the file, or they can download it on the portal) 
-  Sharer downloads signed MTA (with requester signatures) and also signs it.  

Yes, I agree we'll need to add that message asking both Requesters and Reviewers ensure that it signed in all places. We maybe need provide some mechanism to mark that they are missing signatures if the MTA isn't signed correctly.

resources_portal/models/organization_user_permission.py
```diff
@@ -0,0 +1,22 @@
...
+    objects = models.Manager()
+
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    user = models.ForeignKey(User, blank=False, null=False, on_delete=models.CASCADE)
+    organization = models.ForeignKey(
+        Organization, blank=False, null=False, on_delete=models.CASCADE, related_name="permissions"
+    )
+    permission = models.ForeignKey(Permission, blank=False, null=False, on_delete=models.CASCADE)
```
kurtwheeler: I'm not sure the built-in Django permission is going to be able to do what we need. We want to be able to assign users permissions to do specific actions, and these will be organization specific. After a quick glance at the Django permissions, it looks like they allow add, modify, delete permissions for types of objects or specific instances of objects. I don't know that we'll be able to easily encode abstract concepts like "allowed to approve requests for this organization" into that system.

resources_portal/models/material_request.py
kurtwheeler: Ok so the `material_request.mta` field will be for the partially signed MTA that the Requester uploads, and the `material_request.executed_mta` field will be for the fully signed MTA that the Reviewer uploads?

dvenprasad: Can we name `material_request. mta`,  requester_signed_mta for clarity?

kurtwheeler: Wait wouldn't that be `material_request.requester_signed_mta` and `material_request.reviewer_signed_mta` respectively?

resources_portal/models/material_request.py
```diff
@@ -0,0 +1,38 @@
...
+    requester = models.ForeignKey(
+        User, blank=False, null=False, on_delete=models.CASCADE, related_name="material_requests"
+    )
+
+    assigned_to = models.ForeignKey(
+        User, blank=False, null=False, on_delete=models.CASCADE, related_name="assigned_to_set"
+    )
+
+    mta = models.CharField(max_length=255, blank=True, null=True)
+    irb = models.CharField(max_length=255, blank=True, null=True)
```
kurtwheeler: I think both of these fields are supposed to be PDFs.

resources_portal/models/material_request.py
dvenprasad: Yes, but `executed_mta` indicates that all parties have signed and it's ready to go vs `reviewer_signed_mta` which is implies reviewer has signed it and not clear that it is the final state of that pdf.

resources_portal/models/material_request.py
arielsvn: Are we uploading them to the DB with a [FileField](https://docs.djangoproject.com/en/3.0/ref/models/fields/#filefield)? I thought we would upload them to S3 and store the urls here.

resources_portal/models/material_request.py
arielsvn: ok, I renamed the field `mta` to `requester_signed_mta`. Will wait until we clarify how are we storing the pdfs (file field vs url+s3) before updating their types.

resources_portal/models/material_share_event.py
arielsvn: I like it!

resources_portal/models/user_notification.py
arielsvn: Same here, I added it because I saw it on the original sketch. I'm guessing we'll send a notification to users when they need to act on something? Sign, review, etc?

resources_portal/models/material_share_event.py
davidsmejia: wouldn't `material_share_events` be enough? not sure what created adds in terms of context?

arielsvn: yes! even better! :taco:

davidsmejia: also probably should keep owner/user/created_by consistent across models since we have several user properties defined

kurtwheeler: I suggested adding `_created` because there are two foreign relationships between `users` and `material_share_events` so it seems less clear that one of them was created by the user as opposed to being associated with them some other way.

If you both think it's obvious we can drop the `_created`.

resources_portal/models/user_notification.py
kurtwheeler: Seems like something we should figure out then! My guess is that we were going to have this table be for notification settings. For each type of event that we'd notify about we would have a `event_type` choice for. Each user can choose whether or not they want notifications for that type of event for each organization. We can create a default set for each user for each organization they create/join, then delete some of them or create more as the user tweaks their preferences on a per-organization basis. I think that's why there was an `(Org)` written in front of the table name.

I think the assigned_to field was about whether they wanted notifications for all events of that type or just the ones where they were the assignee.

Does that sound right? CC @davidsmejia @dvenprasad 

Also @dvenprasad do you have a list of notification `event_type`s we can use for the choices?

resources_portal/models/material.py
arielsvn: Added the following choices:

```
    MATERIAL_TYPE_CHOICES = (
        ("CELL_LINE", "CELL_LINE"),
        ("PLASMID", "PLASMID"),
        ("PROTOCOL", "PROTOCOL"),
        ("DATASET", "DATASET"),
        ("MOUSE_MODEL", "MOUSE_MODEL"),
        ("ZEBRAFISH_MODEL", "ZEBRAFISH_MODEL"),
    )
```

resources_portal/models/material_request.py
kurtwheeler: Hmm, I like that idea more! Good thought! Can you add a comment or maybe just change the field names to `mta_s3_url` and `irb_s3_url`?

arielsvn: In that case, do we need to have `s3_bucket` and `s3_key` for each one of these fields? These are the arguments required by [S3.upload_file](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.upload_file).

kurtwheeler: Hmm, the alternative would be to store the HTTPS address and then just create a wrapper function for `S3.upload_file` that parses out the bucket and key from the URL. Then we only need one field.

I'm fine with doing it either way, but I think it might make it easier to use our API if just stores/returns a URL.

arielsvn: I like that, let's do it for all files. I'm also going to use a `TextField` for all URLs.

resources_portal/models/user_notification.py
dvenprasad: Yep that sounds about right from what I can recall. 

I don't have a full list of all the `event_types`. Here are the notifications we'll need to send as far as requests are concerned. 
- New Resource Request
- Change in Request status
- Request Approved/Rejected
- Assigned to a request
- Reminders to take actions: Review a request, sign and upload MTAs

Are these sufficient for now? 
There will be more as we dive into each of the modules.

resources_portal/models/material_request.py
cgreene: Is there a reason not to use a URL field?

resources_portal/models/user_notification.py
kurtwheeler: Yeah that's great! Thanks Deepa!

@arielsvn, can you add an Organization to this table, and then also add those `event_types` as choices?

resources_portal/models/material_request.py
kurtwheeler: URLfields are Charfields under the hood, so it needs to have some kind of max length. I think all it gets us is validation by a URLValidator, which I don't really think we need since we'll be creating and setting these S3 urls. I I think I'd rather not deal with an issue in the future where someone is trying to upload a really long filename and we're hitting the maxlength.

cgreene: Is there any security risk with allowing arbitrary filenames and then storing those? Would it be better to assign some sort of uuid or similar to uploaded files?

resources_portal/models/user_notification.py
davidsmejia: should we have a cancelled state?

resources_portal/models/material_request.py
kurtwheeler: Mmm, I think it depends on how we do things: https://owasp.org/www-community/vulnerabilities/Unrestricted_File_Upload

As far as I can tell, if we have the frontend actually just upload the file directly to S3 then we should be fine. I don't really care whether we let the user control the name or not, but I think it should be contextual and not a UUID. For example if we wanted to we could have templates like `material_request_{ID}_requester_signed_MTA.pdf` and `material_request_{ID}_executed_MTA.pdf`.

resources_portal/models/organization_user_permission.py
arielsvn: I think those are the default permissions. I have never tried it but maybe something like [this](https://docs.djangoproject.com/en/3.0/topics/auth/default/#programmatically-creating-permissions) could work:

```py
content_type = ContentType.objects.get_for_model(Organization)
permission = Permission.objects.create(
    codename='approve_requests',
    name='Can approve requests',
    content_type=content_type,
)
```

And then check if a user has that permission with [has_perm](https://docs.djangoproject.com/en/3.0/ref/contrib/auth/#django.contrib.auth.models.User.has_perm):

```
user.has_perm('resources_portal.approve_requests')
```

Do you think something like that could work?

kurtwheeler: I guess my issue is, how does `user.has_perm('resources_portal.approve_requests')` know for which orgs the user has that perm?

arielsvn: good point, hmm... that's logic that we will have to include in any case right? when checking if a user has permissions to approve requests for a specific organization? I was thinking something like:

```
user.has_perm('resources_portal.approve_requests') and \
OrganizationUserPermission.objects.filter(
  user=user, 
  permission__codename="approve_requests", 
  organization=some_org
).exists()
```

It looks like there's some redundant info there. Even if we had our own `Permission` models we would have to do something similar right?

kurtwheeler:
```
OrganizationUserPermission.objects.filter(
  user=user, 
  permission__codename="approve_requests", 
  organization=some_org
).exists()
```
Seems like what I thinking we'd need. What does `user.has_perm('resources_portal.approve_requests')` add on top of that?

arielsvn: yeah, it doesn't add much. Why would we want to roll our version of `Permission`? The model we had is very similar to the one provided by Django which as a bonus integrates with the admin interface.

Also, I just found `django-guardian`, which addresses [this exact problem](https://django-guardian.readthedocs.io/en/stable/userguide/assign.html#for-user). Would this work?

resources_portal/models/user_notification.py
arielsvn: @kurtwheeler I'm having issues understanding this model. Why should we have a different row for each setting instead of a single one for all of them? We could also have a model like `OrganizationUserNotificationSettings`, and add one field for each type of setting. Example:

```py
new_resource_request = models.BooleanField(default=True)
change_request_status = models.BooleanField(default=False)
assignments = models.BooleanField(default=True)
# ... etc
```

@dvenprasad do you have any designs, even lo-fi, of how the users will be able to modify and view these notifications?

resources_portal/models/organization_user_permission.py
arielsvn: We talked and decided to try `django-guardian`, which means we'll no longer need `OrganizationUserPermission`. I'll also add some tests to show how the permissions work both with a specific `Permission` and `Group`.

resources_portal/models/user_notification.py
arielsvn: We talked and decided to rename the model to `OrganizationUserSettings` to be able to store more general settings. I'll leave a comment on the model describing how to add the settings, we can add them once we start working on that feature.

resources_portal/models/organization_user_permission.py
arielsvn: I think it'll work very well, added a test with the following:

```
        user = UserFactory.create()
        organization = OrganizationFactory.create()

        self.assertFalse(user.has_perm("approve_requests", organization))

        self.assertTrue(organization.owner.has_perm("approve_requests", organization))
```

Used a `post_save` signal on `Organization` to assign the default permissions to the organization's owner. We can do something similar with the other permissions that we use.

arielsvn:
@kurtwheeler @davidsmejia this should be ready for another review, take a close look at the permissions in `organization.py`.

dvenprasad:
@kurtwheeler https://github.com/AlexsLemonade/resources-portal/issues/12 is for managing account setting and notifications. Would be nice to track the options on that ticket.

PR #41, opened 2020-03-17
Move backend into api/ directory

kurtwheeler:
## Issue Number

N/A but builds off of #30 

## Purpose/Implementation Notes

This organizes the project a little bit, creating a specific directory for the backend django project. I also added some documentation to the README that I should have done earlier.

## Types of changes

- Organization
- Documenation

## Functional tests

I ran a deploy to make sure that everything still just worked. It did.

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules

PR #44, opened 2020-03-18
Put serializers next to their views

arielsvn:
## Issue Number

N/A

@davidsmejia @kurtwheeler just got this idea, what do you guys think of putting the serializers next to their views? When I first implemented them in separate files I disliked the amount of jumping between files.

This way we would get the information of the API and the info that is transferred together. And the result looks simpler, check this file for example:

https://github.com/AlexsLemonade/resources-portal/blob/9b8e66335d8a30e385129a1f9749f03ec9088626/resources_portal/views/material.py

Still trying to think of drawbacks of this, can you think of any?

## Types of changes


kurtwheeler:
Looks like the tests aren't too happy with you right now.

arielsvn:
hmm tests failed for similar reasons to https://github.com/AlexsLemonade/resources-portal/pull/41, I wonder what's causing that

arielsvn:
Tests passing now @kurtwheeler :)

PR #71, opened 2020-03-30
Kurtwheeler/psql script

kurtwheeler:
## Issue Number

N/A Phil asked "how do I connect to the DB?"

This is almost the same script as refine.bio. I don't quite love it because the database is just called `postgres` locally, but that can be fixed in another issue. For now we have database connectivity!!!

## Types of changes

## Functional tests

I ran it locally and got a CLI for postgres.

## Checklist


api/run_psql_client.sh
```diff
@@ -0,0 +1,2 @@
+#! /bin/sh
+docker run  --network api_default -it -e PGPASSWORD=mysecretpassword --rm --link api_postgres_1:postgres postgres:9.6.6 psql -h postgres -U postgres -d postgres
```
arielsvn: Is it possible to use `docker-compose` for this? I think `api_postgres_1` might change depending on the computer?

kurtwheeler: Ahhh, yes, awesome idea! I'll do that!

PR #74, opened 2020-03-31
Try tweaking the way the README does directories

kurtwheeler:
I'm not sure I love this way either, but I was tired of having to remove `cd X &&` from every command I copied from the README.

I'd love to hear any ideas anyone has of better ways of specifying directories for commands in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,22 +12,24 @@ Resources Portal. Check out the project's [documentation](http://ccdl.github.io/
 
 ## Local Development
 
+(from the `api` directory)
+
 Start the dev server for local development:
 
 ```bash
-cd api && docker-compose up
+docker-compose up
 ```
 
 Run a command inside the docker container:
 
 ```bash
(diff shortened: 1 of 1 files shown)


arielsvn:
hmm our current `readme` is for the API project, would it be good to move it into that directory? And do the same in the `client/` folder? I imagine both projects will have different instructions to execute them.

davidsmejia:
> hmm our current readme is for the API project, would it be good to move it into that directory? And do the same in the client/ folder? I imagine both projects will have different instructions to execute them.

I second this, I think we could have a general readme that links to the specific readmes in the various directories

kurtwheeler:
That sounds good. I'm doing that in my materials PR because I wanted to add some additional API-specific documentation anyway.

kurtwheeler:
I'm going to close this because #77 addresses this in a better way.

Issue #60, opened 2020-03-26
Running list: requests the API will need to support for the end to end tests


kurtwheeler:
### Context

I've created a set of end to end tests for us to implement: #49

This issue will list all of the various requests the API will need to support for the end to end tests. This may include the same route being used for different purposes in order to highlight special parameter the route may need to support.

These may serve as good fuel for unit tests for API as well.

### Endpoints:
- [x] Create account with grant id for personal account
- [x] Create account without grant id
- [ ] List resource on personal organization
- [ ] List resource on non-personal organization
- [ ] Create non-personal organization with grant id
- [ ] Invite user to an organization (see https://github.com/AlexsLemonade/resources-portal/issues/53)
- [ ] Accept invitation to join organization.
- [ ] Grant permissions to organization member
- [ ] Remove permissions from organization member
- [ ] Change notification settings.
- [ ] Search resources
- [ ] Request a resource
- [ ] Approve a request
- [ ] Reviewer marks request fulfilled (see https://github.com/AlexsLemonade/resources-portal/issues/55)
- [ ] Requester marks request fulfilled (see https://github.com/AlexsLemonade/resources-portal/issues/55)

kurtwheeler:
I think this issue has been well implemented! We no longer need this to track things.

PR #77, opened 2020-03-31
Materials route

kurtwheeler:
## Issue Number

#60 

This gets the `/materials` route working and adds unit tests to it.

This also reorganizes the README into three READMEs.

I also renamed `material_type` to `category`.

This is still missing validation and authentication. I'm punting on those until all the routes are working for CRUD operations so the frontend can work off of them ASAP. I created an issue for this: https://github.com/AlexsLemonade/resources-portal/issues/82

## Types of changes

## Functional tests

I tested the route using the `curl` request in the API's README.

## Checklist


kurtwheeler:
Ok @davidsmejia this is no longer a WIP!

api/resources_portal/test/views/test_user.py
```diff
@@ -20,6 +20,9 @@ class TestUserListTestCase(APITestCase):
     def setUp(self):
         self.url = reverse("user-list")
         self.user_data = factory.build(dict, FACTORY_CLASS=UserFactory)
+        # Don't post created_at and updated_at.
```
davidsmejia, line 23: Are we not exposing this just for testing or will we not be able to sort by date attributes on the front end too?

kurtwheeler: This data is for POST. `created_at` and `updated` are read only fields, so they'd be ignored for create or update operations anyway. (see https://www.django-rest-framework.org/api-guide/fields/#read_only)

Issue #76, opened 2020-03-31
Create user route should also generate an organization for that user.


kurtwheeler:
### Context

#15 

### Problem or idea

The plan is that when a user signs up, they get an organization by default. Therefore we should add this to the user's create route.

### Solution or next step

Create each new user an organization.

PR #78, opened 2020-03-31
Creating a user also creates an organization.

kurtwheeler:
## Issue Number

#76 

Every user is gonna need an organization! This creates them in a transaction in the create user route so we can never have a user without one.

## Types of changes

## Functional tests

I created a user with a curl request and it created an org for her!

## Checklist

### DIFF
--- a/api/resources_portal/test/test_views.py
+++ b/api/resources_portal/test/test_views.py
@@ -7,7 +7,7 @@ import factory
 from faker import Faker
 from nose.tools import eq_, ok_
 
-from resources_portal.models import User
+from resources_portal.models import Organization, User
 from resources_portal.test.factories import UserFactory
 
 fake = Faker()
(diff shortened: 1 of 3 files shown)


api/resources_portal/test/test_views.py
```diff
@@ -34,6 +34,8 @@ def test_post_request_with_valid_data_succeeds(self):
         eq_(user.username, self.user_data.get("username"))
         ok_(check_password(self.user_data.get("password"), user.password))
 
+        ok_(Organization.objects.get(owner=user))
```
arielsvn: Are we using `nose` helpers? Instead of `self.assert...`?

kurtwheeler: Hey man, you wrote the rest of the test!

kurtwheeler: :P

arielsvn: hahah yes, just noticed it. It looked weird with the underscore after `ok`

kurtwheeler: But uh, that's a great question. Should we? I tend to follow the existing conventions by default but it appears you didn't set that convention on purpose.

arielsvn: yes, this file was also generated by the template, I think I wrote the permission tests with the default methods. I'd lean towards them, but we can update it in a future PR if you want.

kurtwheeler: I switched all the nose asserts over to unittest asserts.

PR #89, opened 2020-04-03
Add a github action to deploy docs after `dev` is updated

### LINKED ISSUE #45: Deploy docs to github pages
We have a `/docs` folder that can be built with MkDocs. We should be able to deploy these to github pages here.

There's a github action for that https://github.com/marketplace/actions/deploy-mkdocs

arielsvn:
## Issue Number

close #45 

This should will the docs to the repository Github's pages, now hosted at https://alexslemonade.github.io/resources-portal/

Since the docs are general for the project if we want we can add some for the frontend as well.

We would need to think about how to organize the docs and what to put in the readmes vs there.

For reference, to make this work I had to [Generate a personal access token (scope repo)](https://github.com/settings/tokens) and [create a secret](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets) in the repository settings with the name PERSONAL_TOKEN and the value of the token generated. Looks like it's the only way to give permission to the action to publish a page.


arielsvn:
Modified the github action to also deploy storybook from the frontend here https://alexslemonade.github.io/resources-portal/storybook/

davidsmejia:
I want this.

PR #90, opened 2020-04-03
Allow running both the frontend and the API with the same docker-compose command

arielsvn:
## Issue Number

N/A

This moves the `docker-compose` file to the root of the project and adds configuration to also run the frontend. I think there're a few advantages with this setup, mainly related to how easy it will make the development.

Now with a single command: `docker-compose up` an API is started at localhost:8000, and the client is started at localhost:7000 pointing to the local API endpoint. Both have hot reloading, so changing the code updates the pages instantly.

In addition to that, I also made it so that it runs storybook on port 7001. Which is helpful when developing the FE.

@kurtwheeler @davidsmejia This is not yet finished, but I wanted to get your thoughts. The only drawback that I can see is that now the `docker-compose` file initializes 5 services:

- the Postgres database
- the API server at `localhost:8000`
- mkdocs server at `localhost:8001`
- frontend app at `localhost:7000`
- storybook at `localhost:7001`

On my computer it works really well.

If we wanted to move forward with this, it will probably need some changes to the readme and to some of the scripts, because I also renamed `web` to `api` to match the folder name.

## Types of changes

## Functional tests

Tested locally.

## Checklist

## Screenshots

image


davidsmejia:
I think I'm coming around to this idea. I do like the simplicity of dev env. Maybe we should wait until we have a more robust API to dev against?

arielsvn:
Why wait? Do you see any drawbacks to merging it sooner? it should not limit us from anything I think.

davidsmejia:
> Why wait? Do you see any drawbacks to merging it sooner? it should not limit us from anything I think.

Ah ok, I was thinking that the intent was to move away from using `yarn dev` but this is another alternative to run the project as a complete system locally. 

I'm cool with that!

Issue #36, opened 2020-03-17
Component Library Showcase


davidsmejia:
### Context

We want to see how he components behave and are implemented.

### Problem or idea

Add a page to the frontend client that lays out (single file) how the components look

### Solution or next step

after #35 
- [ ]  add [evergreen](https://evergreen.segment.com/components) 
- [ ]  add `/forms-demo`

arielsvn:
Looks like evergreen supports tree-shaking https://github.com/segmentio/evergreen/issues/458#issuecomment-442231720, although it might need a bit more configuration for the icons package.

davidsmejia:
Moved to Grommet as it comes with more components that we would need, though I am thinking it might make sense to use this just as an abstraction for forms.

davidsmejia:
closing with #72

PR #102, opened 2020-04-08
color palette and buttons

davidsmejia:
## Issue Number

#36 + Buttons

This implements the color palette for the resources portal theme in grommet.

An example for how to extend basic components can be seen in the buttons section of `theme.js`.

Some rules begin with `;` which is a ways of avoiding the css parser from doing things like 
`false false color: #00000` because the previous rule is not applied, the false-y value is still printed because... javascript. I'm still considering another cleaner solution but for the moment this will suffice.

Also I would like to implement the attribute of active to  mean something like processing/async activity so we should add some styles that express this, not when the button is being pressed.

## Types of changes

## Functional tests

Check the stories!

## Checklist

- [X] I have added tests that prove my fix is effective or that my feature works

## Screenshots

Screen Shot 2020-04-07 at 11 12 21 PM
Screen Shot 2020-04-07 at 11 12 17 PM


davidsmejia:
Also @dvenprasad there was box shadow on disabled primary buttons was this an oversight, I haven't seen any disabled primary buttons in the comps yet but it seemed anomalous

dvenprasad:
You are correct, there shouldn't be a shadow on disabled buttons. I fixed it on the styleguide and mockups. (There are a few in the `List Resources` module)

client/src/theme.js
```diff
@@ -6,16 +8,69 @@ const theme = {
...
-      ${
-        !props.primary &&
+    extend: (props) =>
+      !props.role &&
+      `
+        ${
+          !props.primary &&
+          !props.disabled &&
+          `
+          ;
```
arielsvn, line 23: This is really cool! But it looks like it's generating invalid CSS, not sure if it's related with these semicolons? Note the final rules for this class, associated with the primary buttons:

```
.hwGyGg{
  display:inline-block;
  box-sizing:border-box;
  cursor:pointer;
  outline:none;
  font:inherit;
  -webkit-text-decoration:none;
  text-decoration:none;
  margin:0;
  background:transparent;
  overflow:visible;
  text-transform:none;
  border:1px solid #017FA3;
  border-radius:4px;
  color:#000000;
  padding:7px 24px;
  font-size:16px;
  line-height:24px;
  background:#017FA3;
  color:#FDFDFD;
  border-radius:4px;
  -webkit-transition-property:color,background-color,border-color,box-shadow;
  transition-property:color,background-color,border-color,box-shadow;
  -webkit-transition-duration:0.1s;
  transition-duration:0.1s;
  -webkit-transition-timing-function:ease-in-out;
  transition-timing-function:ease-in-out;

  false:false;
  color:#FDFDFD;
  -webkit-undefined;
  -ms-flex-undefined;
  undefined;
}
```

Issue #104, opened 2020-04-08
Alerts


davidsmejia:
### Context

Add default styles for alerts and notifications.

https://projects.invisionapp.com/d/main#/console/19722871/412506283/preview

PR #105, opened 2020-04-09
Alerts + Icons + Badges

davidsmejia:
## Issue Number

#104 

Adds alerts as a component with a main alert pipeline/queue.
Still needs to be added to the header but since we don't implement alerts yet I think it's ok to wait. You can checkout the stories to see how it is used and how it will behave.

Also you can pass `<Alerts lifo={false} />` and it will use fifo to preset the stack of alerts!

@dvenprasad svgs are great but the bounding box/viewbox of the check was a little extra which was causing some alignment issues. The icon should be 24px x 24px but it is 32px. The files are `check-white.svg` and `check-green.svg`
```javascript
// you can pass an optional queue parameter if you want to manage a different alert queue
const alertsQueue = useAlertsQueue()
// somewhere else in the code
const { addAlert } = useAlertsQueue()

// Where you want to show the alerts (this component is for convenience which just wraps an Alerts component
<AlertsWithQueue queue={alertsQueue} />
// a callback or button to push an alert to that queue
// the default type is info, supported types are: info, success, error
addAlert('The snozzberries taste like snozzberries!', 'success')
```
## Types of changes

## Functional tests

None yet, will update.

## Checklist

## Screenshots
Screen Shot 2020-04-09 at 7 40 47 PM

Screen Shot 2020-04-09 at 7 39 28 PM


davidsmejia:
Maybe we can remove the linting of SVG files for trailing whitespace?

arielsvn:
> Maybe we can remove the linting of SVG files for trailing whitespace?

Sure, that should work if we add this to the top of the pre-commit config file.

```
exclude: '^.*\.svg$'
```

client/src/components/Alert.js
```diff
@@ -0,0 +1,177 @@
...
+  },
+  success: {
+    Icon: Check,
+    background: 'success',
+    multipleBackground: 'success-shade-20',
+    color: 'white'
+  }
+}
+
+const AlertsContext = React.createContext({ alerts: {} })
```
arielsvn, line 31: It wasn't necessary to use `AlertsContext.Provider`?

PR #118, opened 2020-04-14
Organization route

kurtwheeler:
## Issue Number

#60 

This adds the route for organizations. This ended up being fairly clean and straight forward once I figured everything out. I even got the factory library to work nicely!

I think this took me longer than I expected but I think it will serve as a good pattern for the rest of the resources!

I smushed the migrations into a single one because we don't have a permanent DB up so we don't have to worry about keeping the migrations and I had drifted a bit from `dev` so it seemed like a good enough excuse to smush them down.

## Types of changes

## Functional tests

I'm mostly relying on the unit tests at the moment.

## Checklist


api/resources_portal/models/associations/organization_user_association.py
```diff
@@ -0,0 +1,16 @@
...
+from resources_portal.models.user import User
+
+
+class OrganizationUserAssociation(models.Model):
+
+    organization = models.ForeignKey(
+        Organization, blank=False, null=False, on_delete=models.CASCADE
+    )
+    user = models.ForeignKey(User, blank=False, null=False, on_delete=models.CASCADE)
+
```
davidsmejia, line 13: do we want to add timestamps for associations so we can get metrics for when a user joined an organization?

api/resources_portal/views/organization.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def update_members(self, organization, members):
+        if not members:
+            # Nothing to do.
+            return organization
+
+        member_ids = [member["id"] for member in members]
+
+        organization.members.set(User.objects.filter(id__in=member_ids))
+        organization.save()
```
davidsmejia, line 32: should we return the updated model since we are silently not adding invalid members?

api/resources_portal/models/associations/organization_user_association.py
kurtwheeler: That's a good idea, I'll add a created_at but no updated_at since there's nothing to be updated.

kurtwheeler:
Idk where your comment went. Now that I've addressed it I want to comment on it but Github is hiding it from me. Why?!?

Anyway here's my response to the comment that I addressed but is now missing:

We should be returning the updated organization!!

However we don't silently not add invalid members, the validator will catch that: `{'owner': {'id': ['Must be a valid UUID.']}, 'members': [{}, {}, {'id': ['Must be a valid UUID.']}]}`

I did add a test to verify that adding an invalid member returns a 400.

davidsmejia:
@kurtwheeler yeah i removed that comment when i saw where we called `update_members` from - my b

kurtwheeler:
Ohhh ok, I thought it was further deterioration of Github's API.

kurtwheeler:
> I was also wondering what we would do about managing if owner was in memebers or not, I didn't see anything that handled that specifically.

I think I need to address this in a validator. I'll do that tomorrow.

api/resources_portal/views/organization.py
```diff
@@ -21,13 +21,16 @@ class OrganizationDetailSerializer(OrganizationSerializer):
     owner = UserRelationSerializer()
     members = UserRelationSerializer(many=True)
 
-    def update_members(self, organization, members):
+    def update_members(self, organization, owner_id, members):
         if not members:
```
davidsmejia, line 25: if there are no members we would still want to make sure that the owner is in there on create?

davidsmejia: awesome!

davidsmejia:
LGTM!
Great job!

Issue #122, opened 2020-04-15
Form Header Rows


davidsmejia:
### Context

The forms have an element where the sub section's title is colored like the brand with a trailing line that fills up the rest of the space

https://projects.invisionapp.com/d/main?origin=v7#/console/19276522/402471354/inspect?scrollOffset=0

### Problem or idea

we should add it as a component

### Solution or next step

add it as a component

PR #123, opened 2020-04-15
HeaderRow component

davidsmejia:
## Issue Number

#122

Adds HeaderRow components

## Types of changes

## Functional tests

Story

## Checklist

- [X] I have added tests that prove my fix is effective or that my feature works
- [X] Any dependent changes have been merged and published in downstream modules

## Screenshots

Screen Shot 2020-04-15 at 4 12 17 PM

### DIFF
--- a/client/src/theme.js
+++ b/client/src/theme.js
@@ -731,27 +731,27 @@ const theme = {
   paragraph: {
     large: {
       height: '19px',
-      maxWidth: '235px',
+      maxWidth: '100%',
       size: '15px'
     },
     medium: {
       height: '24px',
-      maxWidth: '192px',
+      maxWidth: '100%',
       size: '16px'
     },
     small: {
       height: '16px',
-      maxWidth: '171px',
+      maxWidth: '100%',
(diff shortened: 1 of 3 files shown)


client/src/theme.js
```diff
@@ -731,27 +731,27 @@ const theme = {
   paragraph: {
     large: {
       height: '19px',
-      maxWidth: '235px',
+      maxWidth: '100%',
```
arielsvn, line 734: does it has the same effect if we remove this value?

davidsmejia: I wanted to keep it there for now as a reminder it exists

davidsmejia: But i think it does because it would be undefined and the default behavior for a `p` would be to expand to fill it's container

davidsmejia: well, expand as it's content dictates until wrapping after filling the container

davidsmejia: set it to undefined to remove the extra rule applied

Issue #112, opened 2020-04-10
Attachments


davidsmejia:
### Context

We are going to want to be able to reuse uploaded items. And manage their uploaded state/deleted state on s3. So we should probably move s3 urls to an attachment model with those attributes.

### Problem or idea

We want to create models to handle the attachments + links to s3.

### Solution or next step

Create an attachment model and replace s3 links with a relationship to these models.

PR #149, opened 2020-04-29
Adds Attachment model.

kurtwheeler:
## Issue Number

#112 

This adds the attachment model and links it up. Wanted to get some eyes on it before i go and change all the test/dev data.

## Types of changes

## Functional tests

I made a migration, but that's about it so far.

## Checklist


kurtwheeler:
Ok @davidsmejia this is no longer a WIP. I've also asked @Philinphiladelphia to take a look!

api/resources_portal/models/attachment.py
```diff
@@ -0,0 +1,37 @@
...
+    objects = models.Manager()
+
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    filename = models.TextField(blank=False, null=True, help_text="The name of the attachment.")
+
+    s3_bucket = models.CharField(max_length=255, blank=True, null=True)
+    s3_key = models.CharField(max_length=255, blank=True, null=True)
+
```
davidsmejia, line 18: It might still be helpful to see an attachment description actually. Unless people are meticulous with their filenames as descriptors... It's not in the designs but I'll @dvenprasad for input.

api/resources_portal/management/commands/populate_test_database.py
```diff
@@ -21,6 +23,11 @@ def handle(self, *args, **options):
         def add_class_to_database(class_json, Class):
             print(f"Inserting {Class.__name__} into database...")
             for element in class_json:
+                if "updated_at" in element:
+                    element["updated_at"] = dateparse.parse_datetime(element["updated_at"])
+                if "created_at" in element:
+                    element["created_at"] = dateparse.parse_datetime(element["created_at"])
```
Philinphiladelphia, line 29: Nice modification!

api/resources_portal/models/attachment.py
```diff
@@ -0,0 +1,37 @@
+from django.db import models
+
+
+class Attachment(models.Model):
+    class Meta:
+        db_table = "attachments"
+        get_latest_by = "created_at"
```
Philinphiladelphia, line 7: I believe Deepa mentioned that we should order/get latest by updated_at.

api/resources_portal/models/attachment.py
```diff
@@ -0,0 +1,37 @@
...
+    s3_bucket = models.CharField(max_length=255, blank=True, null=True)
+    s3_key = models.CharField(max_length=255, blank=True, null=True)
+
+    sequence_map_for = models.ForeignKey(
+        "resources_portal.Material",
+        blank=False,
+        null=True,
+        on_delete=models.SET_NULL,
+        related_name="sequence_maps",
+        help_text="The cell line this seq_map is for. Only valid for seq_map attachments.",
```
Philinphiladelphia, line 25: Should attachment have another field called "type" or something similar which specifies the kind of attachment it is? Might make it easier to process on the front-end.

davidsmejia: I think we will want to add description instead. We should generally try to avoid getting attachments directly and only be getting them from some sort of association to what they are attached to.

kurtwheeler:
@Philinphiladelphia and @davidsmejia I think I addressed your feedback.

api/resources_portal/models/attachment.py
davidsmejia: not for mvp

Issue #135, opened 2020-04-21
Shipping Information & Requirements


davidsmejia:
### Context

We need to list out all of the requirements for shipping (accepted carriers, who pays, etc) for the material.

![Screen Shot 2020-04-21 at 5 42 33 PM](https://user-images.githubusercontent.com/1075609/79916903-846b3280-83f7-11ea-8db8-12f86ab8d2ad.png)

### Problem or idea

The shipping requirements are going to have to be associated with the materials model.
Would be good to include this or just a boolean saying they exist in the search results.

### Solution or next step

Going to talk about it 4/22

PR #150, opened 2020-05-01
Shipping information model

kurtwheeler:
## Issue Number

#135 

Adds the model, migrations, and elasticsearch config for shipping information.

## Types of changes

## Functional tests

I can populate the database with dev data and unit tests are passing.

## Checklist

- [x] Any dependent changes have been merged and published in downstream module


davidsmejia:
After reviewing.... I think we should be calling these `shipping_requirements`. It seems to make more sense as it is displayed for example:

Screen Shot 2020-05-01 at 2 53 58 PM

api/resources_portal/models/shipping_information.py
```diff
@@ -0,0 +1,24 @@
...
+
+    objects = models.Manager()
+
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    needs_shipping_address = models.BooleanField(default=False)
+    sharer_pays = models.BooleanField(default=False)
+
+    ups_code_accepted = models.BooleanField(default=False)
```
davidsmejia, line 17: We should probably use the same semantics for `[adjective]_[noun]` or `[noun]_[adjective]`
So either of the following:
* `needs_shipping_address` and `accepts_ups_code`
* `shipping_address_needed` and `ups_code_accepted`

davidsmejia: Personally, I like `[verb]_[noun]` for truthy values and `[noun]_[adjective]` for relationships and `[verb]_[noun]_[preposition]` for non-truthy vals

api/resources_portal/models/shipping_information.py
```diff
@@ -0,0 +1,24 @@
+from django.db import models
+
+
+class ShippingInformation(models.Model):
+    class Meta:
+        db_table = "shipping_information"
+        get_latest_by = "created_at"
```
Philinphiladelphia, line 7: updated_at?

api/dev_data/shipping_information.json
```diff
@@ -0,0 +1,28 @@
+{
+  "shipping_information": [
+      {
+          "id": "1",
+          "needs_shipping_address": true,
+          "sharer_pays": true,
+          "ups_code_accepted": true,
+          "fedex_code_accepted": true,
+          "reimbursement_accepted": false,
+          "other_payment_methods_accepted": false,
```
Philinphiladelphia, line 10: How will payment work with other shipping methods? Will the requester just have to reach out to the owner to find out what the method is? Should there be a nullable text field to store a "other" shipping method?

kurtwheeler: My assumption is that they would put something into the `restrictions` field? However yes, the sharing of address and/or other payment details will be done via email for the MVP.

davidsmejia: I think this is just to allow people to get an idea of what to expect when they reach out. And "other" is planned on being mostly unused. They can put any other restrictions details in the text box in the form.

api/resources_portal/models/shipping_information.py
kurtwheeler: I like `accepts_ups_code` more but I'm not sure I agree with your general scheme. For example, `created_at` and `updated_at` are `[verb]_[preposition]` and I can't think of a `[verb]_[noun]` that would be more clear than `sharer_pays`. (Although `sharer_pays_for_shipping` might be more clear.)

davidsmejia: Yeah i agree, but I think we could also describe `sharer_pays` as `requires_payment` since requirements are for the requester

api/resources_portal/migrations/0005_auto_20200501_2022.py
```diff
@@ -0,0 +1,46 @@
...
+                    "id",
+                    models.AutoField(
+                        auto_created=True, primary_key=True, serialize=False, verbose_name="ID"
+                    ),
+                ),
+                ("created_at", models.DateTimeField(auto_now_add=True)),
+                ("updated_at", models.DateTimeField(auto_now=True)),
+                ("needs_shipping_address", models.BooleanField(default=False)),
+                ("sharer_pays", models.BooleanField(default=False)),
+                ("accepts_ups_code", models.BooleanField(default=False)),
```
davidsmejia, line 27: hey so `accepts_ups_code` and `accepts_fedex_code` should be `accepts_shipping_code` and then we will tell them to clarify in the restrictions.

kurtwheeler:
@davidsmejia how you like them fields?

PR #151, opened 2020-05-01
Fix the remaining date fields that have issues.

kurtwheeler:
## Issue Number

I fixed most of these fields in #149

@davidsmejia pointed out that embargo_date was still a TextField. I think after this everything is well though!

## Types of changes

## Functional tests

Populating the database with dev data works smoothly!

## Checklist

### DIFF
--- a/api/resources_portal/management/commands/populate_test_database.py
+++ b/api/resources_portal/management/commands/populate_test_database.py
@@ -27,6 +27,8 @@ class Command(BaseCommand):
                     element["updated_at"] = dateparse.parse_datetime(element["updated_at"])
                 if "created_at" in element:
                     element["created_at"] = dateparse.parse_datetime(element["created_at"])
+                if "date_joined" in element:
+                    element["date_joined"] = dateparse.parse_datetime(element["date_joined"])
 
                 element_in_class = Class(**element)
                 element_in_class.save()
--- a/api/resources_portal/models/documents.py
+++ b/api/resources_portal/models/documents.py
@@ -90,7 +90,7 @@ class MaterialDocument(Document):
     pre_print_title = fields.TextField()
     citation = fields.TextField()
     additional_info = fields.TextField()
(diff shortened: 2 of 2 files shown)


api/resources_portal/management/commands/populate_test_database.py
```diff
@@ -27,6 +27,8 @@ def add_class_to_database(class_json, Class):
                     element["updated_at"] = dateparse.parse_datetime(element["updated_at"])
                 if "created_at" in element:
                     element["created_at"] = dateparse.parse_datetime(element["created_at"])
+                if "date_joined" in element:
+                    element["date_joined"] = dateparse.parse_datetime(element["date_joined"])
```
davidsmejia, line 31:
```suggestion
                    element["date_joined"] = dateparse.parse_datetime(element["date_joined"])
                if "embargo_date" in element:
                    element["embargo_date"] = dateparse.parse_datetime(element["embargo_date"])
```

api/resources_portal/management/commands/populate_test_database.py
```diff
@@ -27,6 +27,8 @@ def add_class_to_database(class_json, Class):
...
```
davidsmejia, line 32:
```suggestion
                if "embargo_date" in element:
                    element["embargo_date"] = dateparse.parse_datetime(element["embargo_date"])
```

kurtwheeler:
> Looks good maybe we want to also parse embargo_date?

It is parsed! However because its a DateField instead of a DateTimeField the default parser doesn't trigger a warning about a timezone naive datetime.

api/resources_portal/management/commands/populate_test_database.py
kurtwheeler: This isn't necessary, see: https://github.com/AlexsLemonade/resources-portal/pull/151#issuecomment-622525988

Issue #53, opened 2020-03-26
How will members join organizations?


kurtwheeler:
### Context

Our data model has organizations that will have an owner and then other members.

### Question

How will a member be added to an organization?  Does a user get a invitation or do they request to be allowed in? I guess depending which way it goes we' also need to allow accepting an invitation or approving a request?

Tagging @dvenprasad for an answer.

dvenprasad:
We should allow both invites and request to join, **but for MVP, let's do invites only**. 

For requesting to join a lab, we'll need to provide a mechanism to find people on the portal and their labs, like a member directory or something.... I don't think this is super important for MVP.

kurtwheeler:
Ok, so then we probably need to have a OrganizationInvitations table so we can track these. I'm thinking we can use one table for both directions, with a field that indicates whether it's a invitation or a request.

Initial fields for this model:
```
requester: User
request_receiver: User
organization: Organization
invite_or_request: String (choices: INVITE or REQUEST) (default INVITE)
status: String (choices: PENDING, ACCEPTED, REJECTED) (default: PENDING)
created_at: datetime
updated_at: datetime
```
Basic idea is that the `requester` sends an OrganizationInvitation to the `request_receiver` to join `organization`, unless `request_or_invite == REQUEST`, in which case the requester is asking to join `organization` instead. The status is updated when the request is accepted or rejected by the `request_receiver`.

davidsmejia:
I was thinking of potential issues after discussing with @kurtwheeler, for after the MVP, it might make sense to add an additional status of `INVALID` or `EXPIRED` to account for requests made by a previous owner of an organization whom no longer has perms to modify the members at time of request being accepted (or would have been accepted).

PR #152, opened 2020-05-01
Add invitations route

Philinphiladelphia:
## Issue Number

#53

This adds the organization invitations route. It also checks user's authentication and permissions on this route.

## Functional tests

I added tests for each of the routes and permissions, and manually tried out the request types.

## Checklist

## Screenshots


api/resources_portal/models/user.py
```diff
@@ -16,6 +16,10 @@ class User(AbstractUser):
 
     organizations = models.ManyToManyField("Organization", through="OrganizationUserAssociation")
 
+    def save(self, *args, **kwargs):
+        self.id = uuid.UUID(str(self.id))
+        super(User, self).save(*args, **kwargs)
```
kurtwheeler, line 21: Did this end up being necessary? I think I was pairing with you while you wrote this and we did it to address a red herring IRRC. Can you try taking it out and seeing if everything still works?

api/resources_portal/views/organization_invitations.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class OrganizationInvitationSerializer(serializers.ModelSerializer):
+    requester = serializers.PrimaryKeyRelatedField(queryset=User.objects.all())
+    request_reciever = serializers.PrimaryKeyRelatedField(queryset=User.objects.all())
+    organization = serializers.PrimaryKeyRelatedField(queryset=Organization.objects.all())
+
+    def validate(self, data):
+        if not data["request_reciever"].has_perm(
+            "add_members_and_manage_permissions", data["organization"]
```
kurtwheeler, line 15: I don't think this quite covers the nuances of the permissions necessary for this route. The flow will be:
1. `requester` POSTs an invitation (permission here is a bit complex because anyone may make a `REQUEST`, but only someone with permission can create an `INVITE`).
2. `request_receiver` is the only one who may PUT to update an existing request by approving or declining it. (Although note that the `requester` is the only one who should be able to DELETE the invitation.)
3. Once an invitation has been approved or denied, it should no longer be mutable.

kurtwheeler: Also I think this needs to be done in the view because you need to see who `request.user` is because you cannot assume `data['request_reciever']` is the user that made the HTTP request.

api/resources_portal/models/user.py
Philinphiladelphia: Just tested this- it still seems to be neccesary.

api/resources_portal/views/organization_invitations.py
Philinphiladelphia: Alright, I think the proper permissions are in place. Have a look at your leisure.

api/resources_portal/views/organization_invitations.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def create(self, request, *args, **kwargs):
+        if not request.user.is_authenticated:
+            return Response(status=status.HTTP_403_FORBIDDEN)
+
+        if not OrganizationInvitationSerializer(data=request.data).is_valid():
+            return Response(status=status.HTTP_400_BAD_REQUEST)
+
+        request_reciever = User.objects.get(pk=request.data["request_reciever"])
+        organization = Organization.objects.get(pk=request.data["organization"])
```
kurtwheeler, line 50:
```suggestion
        serializer = OrganizationInvitationSerializer(data=request.data)
        serializer.is_valid(raise_exception=True)

        request_reciever = serializer.validated_data.request_reciever
        organization = serializer.validated_data.organization
```
Source: https://www.django-rest-framework.org/api-guide/serializers/#raising-an-exception-on-invalid-data

Make sure to test that, Github isn't my preferred editor.

api/resources_portal/views/organization_invitations.py
```diff
@@ -0,0 +1,103 @@
...
+        requester_accepting = (
+            request.user == invitation.requester and invitation.invite_or_request == "INVITE"
+        )
+        request_reciever_approving = (
+            request.user == invitation.request_reciever
+            and invitation.invite_or_request == "REQUEST"
+        )
+
+        # Xor operator
+        if not requester_accepting ^ request_reciever_approving:
```
kurtwheeler, line 76:
```suggestion
        if not (requester_accepting or request_reciever_approving):
```
I feel like XOR is generally less clear unless it's important that one and only one condition is true. Both of the conditions you're testing cannot be true at the same time. We're really just testing that at least one of them is true.

api/resources_portal/views/organization_invitations.py
```diff
@@ -0,0 +1,103 @@
...
+            return Response(status=status.HTTP_400_BAD_REQUEST)
+
+        request_reciever = User.objects.get(pk=request.data["request_reciever"])
+        organization = Organization.objects.get(pk=request.data["organization"])
+
+        if request.data["invite_or_request"] == "INVITE" and not request_reciever.has_perm(
+            "add_members_and_manage_permissions", organization
+        ):
+            return Response(
+                f"{request_reciever} does not have permission to add members",
```
kurtwheeler, line 56: Can we return all errors as JSON? This is what one of our existing errors looks like:
```
{"detail":"Method \"GET\" not allowed."}
```

Issue #91, opened 2020-04-03
Implement initial version of the search page with a single resource type


arielsvn:
### Context

https://github.com/AlexsLemonade/resources-portal/issues/63
https://github.com/AlexsLemonade/resources-portal/issues/13

### Problem or idea

Add the search page, high-fi designs on https://github.com/AlexsLemonade/resources-portal/issues/13. For this one let's just:

- Add the search input at the top
- Leave the space for the filters (will file another issue for them)
- Decide how we are going to store the templates for each resource type and add one for GEO

Proposed URL: `/search/?q=mouse`

davidsmejia:
closed with #63 + #153

PR #153, opened 2020-05-04
search page

davidsmejia:
## Issue Number

#91 

This adds in the logic... and some refactors... for search.
This currently uses bogus test data. So we will still need to update the `/api/search` to point to the correct endpoint
But once that is implemented the following should be fully functional:
* pagination ( still want to abstract the component out to be reusable )
* search box/query string with dropdown for resource type
* filtering
* shareable urls for search results
* resource types correctly rendered
* adds empty search case but waiting on designs for that

Additionally: 
* fixes all known console errors (didn't impact experience but was nice to get rid of)
* moves resources components to components dir getting ready for resources refactor
* generalizes search hooks so other search endpoints should be as simple as pointing them to the correct endpoint
* adds application context so we can manage user authentication, search (so we can send them back to where they left off), and now have a place to stash other incomplete forms and a place to hook in that logic

## Types of changes

## Functional tests

Nothing automated yet.

## Checklist

## Screenshots

Screen Shot 2020-05-03 at 11 42 47 PM
Screen Shot 2020-05-03 at 11 42 55 PM


client/.eslintrc.js
```diff
@@ -42,6 +42,7 @@ module.exports = {
     'no-use-before-define': 0,
     'react/no-multi-comp': 0,
     'import/no-mutable-exports': 0,
-    'react/jsx-props-no-spreading': 'warn'
+    'react/jsx-props-no-spreading': 'warn',
+    'no-console': 'warn'
```
davidsmejia, line 46: I will remove these but for now, while we have test data, I think it makes sense to keep for now.

client/src/ResouresPortalContext.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+
+export const ResourcesPortalContext = React.createContext([{}, () => {}])
```
davidsmejia, line 3: This could should probably just be an empty object for now. It is supposed to resemble what the structure of the object will look like but the way the provider is designed it wont really affect anything.

client/src/components/SearchInput.js
```diff
@@ -1,16 +1,95 @@
...
+    addFacet,
+    removeFacet,
+    goToSearchResults,
+    setSearchString
+  } = useMaterialsSearch()
+  const [resourceType, setResourceType] = React.useState('ALL')
+  const [string, setString] = React.useState(search || initialString)
+  const resourceOptions = [
+    { label: 'All', value: 'ALL' },
+    ...sortedObjectKeys(Mappings).map((mapping) => {
```
davidsmejia, line 24: This could be pulled out into a constant so I will be easier to read that you are converting the mapping keys to objects for the select to consume.

client/src/components/SearchInput.js
```diff
@@ -1,16 +1,95 @@
...
+      return {
+        label: getReadable(mapping.key),
+        value: mapping.key
+      }
+    })
+  ]
+
+  const submitForm = () => {
```
davidsmejia, line 32: maybe rename to `handleSubmit` or something similar since there is no explicit concept of form here

client/src/components/SearchResult.js
```diff
@@ -165,32 +175,35 @@ export const RequestRequirements = ({ resource }) => {
 
   const requirements = []
   const MTA = 'Material Transfer Agreement'
+  if (resource.needs_abstract) requirements.push('Abstract')
   if (resource.needs_irb) requirements.push('IRB')
   if (resource.needs_mta) requirements.push(MTA)
-  if (resource.needs_mta) requirements.push('Abstract')
-  // if (resource.needs_shipping) requirements.push('Shipping Information')
+  // if (resource.shipping_information) requirements.push('Shipping Information')
```
davidsmejia, line 181: This is available in the test data now, so you can implement this.

client/src/components/SearchResultsFilters.js
```diff
@@ -0,0 +1,104 @@
+import React from 'react'
+import { Box, Text, Heading, CheckBox } from 'grommet'
+import { getReadable } from '../helpers/readableNames'
+import { sortedObjectKeysByValues } from '../helpers/sortObjectKeys'
+import { useMaterialsSearch } from '../hooks/useMaterialsSearch'
+
+const isChecked = (queryFacets, facet) => {
```
davidsmejia, line 7: The performance of this could go from exp to linear as the facets increases. However as it stands now there won't be so many facets that it would be noticeable.

client/src/components/SearchResultsLimit.js
```diff
@@ -0,0 +1,37 @@
...
+    goToSearchResults
+  } = useMaterialsSearch()
+
+  React.useEffect(() => {
+    if (!limit) setLimit(limitOptions[0])
+  })
+
+  return (
+    <Box direction="row" align="center">
+      Showing
```
davidsmejia, line 21: Maybe this should be wrapped in a tag?

client/src/components/SearchResultsOffset.js
```diff
@@ -0,0 +1,155 @@
...
+    setEnteredPageNumber('')
+    goToSearchResults()
+  }
+
+  const atEnd = safeOffset === lastOffset
+  const atStart = safeOffset === 0
+
+  const handlePageNumberRequest = ({ target: { value } }) => {
+    const parsedValue = parseInt(value, 10)
+    if (/^\d+$/.test(value)) {
```
davidsmejia, line 47: regexs are hard, this tests that only numbers are present maybe make it a variable that explains that

client/src/helpers/sortObjectKeys.js
```diff
@@ -0,0 +1,20 @@
+// transforms an object into an array
+
+export const sortedObjectKeys = (obj, sort) => {
```
davidsmejia, line 3: maybe add a note that explains this is alphabetical order since it isn't explicit

client/src/helpers/sortObjectKeys.js
```diff
@@ -0,0 +1,20 @@
...
+  return Object.keys(obj)
+    .sort(sort)
+    .map((key) => ({
+      key,
+      value: obj[key]
+    }))
+}
+
+export const sortedObjectKeysByValues = (obj, ascending = true) => {
```
davidsmejia, line 12: this might be able to be reworked to handle sorting numerical and alphabetical values like `1-9a-z`

client/src/hooks/useMaterialsSearch.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { useSearch } from './useSearch'
+import { ResourcesPortalContext } from '../ResouresPortalContext'
+
+export const useMaterialsSearch = (newSearch) => {
```
davidsmejia, line 5: rename this to `useResourcePortalSearch` since this uses the `ResourcePortalContext` (to be consistent with other searches that aren't "global")

client/src/hooks/useMaterialsSearch.js
```diff
@@ -0,0 +1,17 @@
...
+  const { search, setSearch } = React.useContext(ResourcesPortalContext)
+  // Should only be populated by results from get initial props
+  if (newSearch) Object.assign(search, newSearch)
+
+  const defaultSearch = useSearch(search, setSearch)
```
davidsmejia, line 10: 😎

client/src/pages/_app.js
```diff
@@ -3,13 +3,14 @@ import { Grommet } from 'grommet'
 import Head from 'next/head'
 import theme from '../theme'
 import Layout from '../components/Layout'
+import { ResourcesPortalProvider } from '../ResouresPortalContext'
```
davidsmejia, line 6: doesn't matter but maybe this sounds better as `ResourcePortalContextProvider`

client/src/styles/app.scss
```diff
@@ -1,5 +1,5 @@
 // global scss file
-@import 'ress';
+// @import 'ress';
```
davidsmejia, line 2: remove the yarn dep

davidsmejia:
Change anything from `ResourcePortal` to `ResourcesPortal`

client/src/hooks/useMaterialsSearch.js
davidsmejia: changed to `useSearchResources`

client/src/hooks/useSearch.js
```diff
@@ -0,0 +1,85 @@
...
+  const router = useRouter()
+
+  const goToSearchResults = (goToStart = false) => {
+    if (goToStart) delete search.query.offset
+    const { pathname, query } = search
+    router.push({ pathname, query })
+  }
+
+  const setSearchString = (searchString, resetQuery = false) => {
+    // should this clear the search facets?
```
kurtwheeler, line 15: I would imagine so right? Couldn't the user could have potentially selected some facets that aren't even in the results of the new search?

davidsmejia: I think it might be more complicated, so I made it an option since it might change depending on context... like if you wanted to use this in a component that searched users that you could filter by organization..
You might want to change the name but keep the organization facet in the search query

client/src/hooks/useSearch.js
```diff
@@ -0,0 +1,85 @@
...
+    if (searchString === '') {
+      delete search.query.search
+    } else {
+      search.query.search = searchString
+    }
+
+    setSearch({ ...search })
+  }
+
+  // get a list of unique items we can add to or delete from
```
kurtwheeler, line 26:
```suggestion
  // get a set of unique items we can add to or delete from
```

client/src/hooks/useSearch.js
kurtwheeler: Makes sense. A default of true might make sense though right?

client/src/hooks/useSearch.js
```diff
@@ -0,0 +1,85 @@
...
+    if (!existingFacet) return new Set()
+    if (Array.isArray(existingFacet)) return new Set(existingFacet)
+    return new Set([existingFacet])
+  }
+
+  const addFacet = (facet, value) => {
+    if (value) {
+      const newFacet = getSetFromFacet(facet)
+      newFacet.add(value)
+      search.query[facet] = [...newFacet]
```
kurtwheeler, line 38: Is it not possible to just update the search object to a set for query instead of a list everywhere rather than converting back and forth?

davidsmejia: it can't be a set unless we transform it to an array before sending the query to the next router. Also when next supplies the query from the route it doesn't know if some query parameters should be arrays or a string:

`/search?category=DATASET`
would have category as `{category: 'DATASET'}`
and 
`/search?category=DATASET&category=MODEL_ORGANISM`
would have category as `{category: ['DATASET', 'MODEL_ORGANISM']}`
also 
`/search?category=DATASET`
can be created by `{category: ['DATASET']}` or {category: 'DATASET'}
but it will always be the latter when supplied from next

Using set just makes this easier but it does need to be switch back to an array at some point, I figured this was nicer as opposed to accessing a query with a getter method instead

client/src/helpers/sortObjectKeys.js
davidsmejia: added a note that it expects all numerical values in the object... maybe we can improve this or add another method to check if the keys are homogeneous or how to handle them

client/src/hooks/useSearch.js
kurtwheeler: Ah word. Yeah this is a nice way to do this. Just wanted to make sure we couldn't just use a set everywhere instead.

dvenprasad:
-  The background on the dropdown button(?) needs to be #f2f2f2 (black-tint-95)
Screen Shot 2020-05-04 at 1 04 34 PM


- I would reduce the width of the dropdown to fit `100`
Screen Shot 2020-05-04 at 1 08 33 PM

- `Publication Info` -> `Publication Information`
- `has` -> `Has` (capitalize the h) 
Screen Shot 2020-05-04 at 1 08 46 PM


-  The selected page and disabled button have the same color. Selected state should have different color. We can use black maybe? #000000
Screen Shot 2020-05-04 at 1 08 55 PM

- Can we add `Jump to Page` as a label instead of a placeholder and reduce the width of the textbox? 
Screen Shot 2020-05-04 at 1 08 58 PM

- Can we add a `Clear all` button to the filters?
Screen Shot 2020-05-04 at 1 20 41 PM

client/src/hooks/useSearchResources.js
```diff
@@ -0,0 +1,17 @@
...
+  if (newSearch) Object.assign(search, newSearch)
+
+  const baseSearch = useSearch(search, setSearch)
+
+  return {
+    ...baseSearch,
+    query: search.query,
+    response: search.response
+  }
+}
```
kurtwheeler, line 17: What's the distinction between this hook and useSearch? As far as I can tell this just uses the context to attach query and response, but nothing is actually specific to Resources? Couldn't that just be done in useSearch as well? Or is this in anticipation that different searches will attach different queries and responses? But still, doesn't it ultimately depend on what the context provided to the hook is?

client/src/hooks/useSearchResources.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { useSearch } from './useSearch'
+import { ResourcesPortalContext } from '../ResouresPortalContext'
```
kurtwheeler, line 3:
```suggestion
import { ResourcesPortalContext } from '../ResourcesPortalContext'
```
This will necessitate a file change as well.

client/src/pages/_app.js
```diff
@@ -3,13 +3,14 @@ import { Grommet } from 'grommet'
 import Head from 'next/head'
 import theme from '../theme'
 import Layout from '../components/Layout'
+import { ResourcesPortalContextProvider } from '../ResouresPortalContext'
```
kurtwheeler, line 6:
```suggestion
import { ResourcesPortalContext } from '../ResourcesPortalContext'
```
This will necessitate a file change as well.

client/src/hooks/useSearchResources.js
davidsmejia: Yeah this is basically the same implementation as any other search type endpoint but this one needs to connect to the application context to preserve the response... other searches can just be regular `useSearch` hooks to interact with setting the search query etc

kurtwheeler:
Did you mean to add empty API files such as `client/src/api/organizations.js`?

client/src/pages/_app.js
davidsmejia: good catch i updated the file and this in a single commit!

davidsmejia:
> Did you mean to add empty API files such as `client/src/api/organizations.js`?

Yeah I was just scaffolding them as they will be populated for now with testData

client/src/components/resources/CellLine/index.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
-import { SearchResult, SearchResultDetail } from '../../components/SearchResult'
-import DetailsTable from '../../components/DetailsTable'
-import { HeaderRow } from '../../components/HeaderRow'
+import { SearchResult, SearchResultDetail } from '../../SearchResult'
+import DetailsTable from '../../DetailsTable'
+import { HeaderRow } from '../../HeaderRow'
```
kurtwheeler, line 4: Just wondering, but does javascript allow absolute imports?

client/src/hooks/useSearchResources.js
kurtwheeler: Got it.

client/src/components/resources/CellLine/index.js
davidsmejia: Honestly I don't know what the default functinality is... the code and this part especially is rewritten by babel and webpack I think there is a way to configure mappings for paths so we could say `import thing from components` and it would know to look for `thing` in `components/thing/index.js` but I haven't had a chance to dig into that yet

client/src/api/search.js
```diff
@@ -0,0 +1,42 @@
...
+    },
+    has_publication: 200,
+    has_pre_print: 12
+  },
+  results: materialsTestData.slice(0, 10)
+}
+
+export const searchResources = async () => {
+  // perform fetch from api
+  console.log('faking materials api search endpoint (2s)')
```
kurtwheeler, line 31:
```suggestion
  console.log('faking materials api search endpoint (20ms)')
```

client/src/common/helpers.js
```diff
@@ -1,266 +0,0 @@
...
-  ]
-  const rx = /\.0+$|(\.[0-9]*[1-9])0+$/
-  let i
-  for (i = si.length - 1; i > 0; i -= 1) {
-    if (num >= si[i].value) {
-      break
-    }
-  }
-  return (num / si[i].value).toFixed(digits).replace(rx, '$1') + si[i].symbol
-}
```
kurtwheeler, line 266: Was this copied from another project and now you've moved each of the functions this project actually uses into its own file?

client/.eslintrc.js
kurtwheeler: Can you make an issue to track that? Seems like an easy thing to forget about.

client/src/theme/index.js
```diff
@@ -272,7 +308,7 @@ const theme = {
         large: '96px',
         medium: '64px',
         small: '48px',
-        xlarge: '144px'
+        xxlarge: '144px'
```
kurtwheeler, line 311: This change is intentional?

client/src/components/SearchInput.js
```diff
@@ -1,16 +1,96 @@
...
+}) {
+  const {
+    query: { search },
+    addFacet,
+    removeFacet,
+    goToSearchResults,
+    setSearchString
+  } = useSearchResources()
+  const [resourceType, setResourceType] = React.useState('ALL')
+  const [string, setString] = React.useState(search || initialString)
```
kurtwheeler, line 21: I don't love `string` as a variable name.

client/src/components/SearchInput.js
```diff
@@ -1,16 +1,96 @@
...
+import { sortedObjectKeys } from '../helpers/sortObjectKeys'
+import { getReadable } from '../helpers/readableNames'
+
+export default function SearchInput({
+  initialString = '',
+  onChange,
+  size = 'medium'
+}) {
+  const {
+    query: { search },
```
kurtwheeler, line 14: Can this variable be `searchString`?

client/src/components/SearchInput.js
kurtwheeler: Could we maybe go with `currentSearchString` or something? It's maybe a bit confusing keeping initialString, string, and search differentiated, so I think more descriptive names might help.

I think I'd also be in favor of changing `initialString` -> `intialSearchString`. I think the concept of "search string" is clear, but if you shorten it down to just `string` it starts to get really ambiguous.

client/src/components/SearchInput.js
```diff
@@ -1,16 +1,96 @@
...
+      </Box>
       <Button
         label="Search"
         size={size}
         width="96px"
         margin={{ left: 'medium' }}
+        onClick={handleSubmit}
       />
     </Box>
   )
```
kurtwheeler, line 96: This file should probably be named `SearchResourcesInput.js` right?

client/src/components/SearchResult.js
```diff
@@ -1,18 +1,23 @@
 import React from 'react'
```
kurtwheeler, line 1: I suppose we can wait to rename all the `Search*` files until there's other types of search results, but this is another file whose name doesn't seem specific enough. Could be `SerachResourcesResult.js`.

davidsmejia: Yeah thats a fair point, the grouping is because this is for the `/search` or global search - it probably should be `ResourceSearchResult.js` so it is consistent with `ResourceDetails` etc

client/src/components/SearchResult.js
```diff
@@ -49,10 +54,15 @@ export const SearchResult = ({ resource, children, hideDefaults = false }) => {
...
               {getReadable(resource.category)}
             </Box>
-            {resource.additional_metadata.organism && (
-              <Box as="span" gap="small" align="center" direction="row">
+            {resource.organism && (
+              <Box as="span" align="center" direction="row" gap="small">
                 <OrganismIcon />
-                {resource.additional_metadata.organism}
+                <span>
+                  {resource.organism.map((organism, i) => [
```
kurtwheeler, line 61: Woah, `organism` is a list. What a twist!!! This should definitely be changed to `organisms`, but I can see that it's an issue even in the Django model.

client/src/components/SearchInput.js
davidsmejia: this is to match the API
https://github.com/AlexsLemonade/resources-portal/blob/dev/api/resources_portal/views/document_views.py#L143

client/src/components/SearchResult.js
```diff
@@ -65,16 +75,15 @@ export const SearchResult = ({ resource, children, hideDefaults = false }) => {
...
-              </Link>
+              <Button
+                as="a"
+                href={resource.url}
+                label={`View on ${
+                  getReadable(resource.import_source) || 'Source Site'
+                }`}
+                margin={{ bottom: 'small' }}
+                primary
+              />
```
kurtwheeler, line 86: Why the change from Link to Button?

client/src/components/SearchResult.js
```diff
@@ -165,32 +180,35 @@ export const RequestRequirements = ({ resource }) => {
...
-            </React.Fragment>
-          ))}
-        </Paragraph>
+        <span>
+          {requirements.map((req, i) => [
+            i !== 0 && ', ',
+            req === MTA ? (
+              <Anchor
+                key={req}
+                href={resource.mta_s3_url}
```
kurtwheeler, line 202: Note that this field has gone away.

client/src/common/helpers.js
davidsmejia: More or less yeah. They were copied from refine bio but I don't think I want to solve all of the problems the same way. I'd like to keep all of the helpers to be small descriptive bits of code that are reusable. Also it will be easier for me to check to see which files aren't being included so we can manage dead code better.

client/src/theme/index.js
davidsmejia: good catch! this should be xlarge

client/src/components/SearchInput.js
davidsmejia: `initialString` -> this actually should be deleted and only be populated from the hook
`string` -> `inputValue`
`setString` -> `setinputValue`

davidsmejia: also changed `search` to `query.search` for clarity

davidsmejia: and also `setSearchString` is now `setSearchTerm` in the hook

client/src/components/SearchInput.js
davidsmejia: maybe `SearchTermInput.js`

client/src/components/SearchInput.js
davidsmejia: i changed it to be obvious that it is for the query object to `query.search` the other option was rewriting it but id rather have it match

client/src/components/SearchResult.js
davidsmejia: yeah we should rename it and when the organisms stop showing up we can switch it then

client/src/components/SearchResult.js
davidsmejia: We just didn't need the wrapper the `as="a"` does the same in terms of rendering but will also supply the button styles as per the designs. Also since this links out the an external url we don't need `next/router` to manage that

client/src/components/SearchResult.js
davidsmejia: good catch I can update after the attachments stuff is in the books

davidsmejia: added a note too

client/src/components/SearchResultsOffset.js
```diff
@@ -0,0 +1,167 @@
...
+  const safeOffset = parseInt(offset, 10) || 0
+  const safeCount = parseInt(count, 10) || 0
+  const safeLimit = parseInt(limit, 10)
+  const lastOffset =
+    safeCount && safeLimit > 0 ? Math.floor(safeCount / safeLimit) : 0
+
+  const startingPageSlices =
+    safeOffset <= 2 ? [0, 6] : [safeOffset - 2, safeOffset + 3]
+
+  const pageSlices = lastOffset - safeOffset <= 2 ? [-5] : startingPageSlices
```
kurtwheeler, line 30: I think these calculations could be written a bit more clearly. I think an if statement would work better here than the ternary operator. `startingPageSlices` doesn't even need to be defined unless `lastOffset - safeOffset` is less than 2. It might be helpful to name `lastOffset - safeOffset` as a variable, because I'm not really sure what that number represents and why it's being compared to 2.

client/src/components/SearchResultsOffset.js
```diff
@@ -0,0 +1,167 @@
...
+
+  const [enteredPageNumber, setEnteredPageNumber] = React.useState('')
+  const [
+    enteredPageNumberInRange,
+    setEnteredPageNumberInRange
+  ] = React.useState(false)
+
+  const safeOffset = parseInt(offset, 10) || 0
+  const safeCount = parseInt(count, 10) || 0
+  const safeLimit = parseInt(limit, 10)
```
kurtwheeler, line 23: It seems odd to me that safeLimit might be NaN. Perhaps `parsedLimit` might be better?

client/src/components/SearchResultsOffset.js
```diff
@@ -0,0 +1,167 @@
...
+    setOffset(page)
+    setEnteredPageNumber('')
+    goToSearchResults()
+  }
+
+  const atEnd = safeOffset === lastOffset
+  const atStart = safeOffset === 0
+
+  const handlePageNumberRequest = ({ target: { value } }) => {
+    const parsedValue = parseInt(value, 10)
```
kurtwheeler, line 46: Seems to make sense to move this down a line, no need to parse it if it's not all numbers.

client/src/components/SearchResultsOffset.js
```diff
@@ -0,0 +1,167 @@
...
+    if (isOnlyNumbers(value)) {
+      const pageOffset = parsedValue - 1
```
kurtwheeler, line 48: This is because pages are 0 indexed but that's confusing for users?

client/src/components/SearchResult.js
kurtwheeler: In which books?

client/src/components/SearchResultsOffset.js
davidsmejia: yeah you expect to see page `1` of results but that corresponds to an offset of `0`

client/src/components/SearchResultsOffset.js
davidsmejia: yeah actually we could get rid of it

client/src/components/SearchResultsOffset.js
```diff
@@ -0,0 +1,166 @@
...
+      const pageOffset = parseInt(value, 10) - 1
+      const inRange = pageOffset <= lastOffset && pageOffset >= 0
+      setEnteredPageNumberInRange(inRange)
+      setEnteredPageNumber(value)
+    }
+    if (value === '') setEnteredPageNumber('')
+  }
+
+  const goToOffsetRequest = () => {
+    goToOffset(parseInt(enteredPageNumber, 10) - 1)
```
kurtwheeler, line 56: Why don't you check `isOnlyNumbers` here as well?

davidsmejia: we only need to make sure that enteredPageNumber is vetted before we set it

client/src/components/SearchResultsOffset.js
```diff
@@ -0,0 +1,166 @@
...
+  const pageSlices = lastOffset - safeOffset <= 2 ? [-5] : startingPageSlices
+
+  const pageButtonsOffsets = [...Array(lastOffset + 1).keys()].slice(
+    ...pageSlices
+  )
+
+  const goToOffset = (page) => {
+    setOffset(page)
+    setEnteredPageNumber('')
+    goToSearchResults()
```
kurtwheeler, line 39: If `page` is the same as the current offset, will `goToSearchResults()` actually do anything or will it recognize that nothing has changed?

client/src/components/SearchResultsOffset.js
kurtwheeler: Ah I see, so it's already been set so it's already been vetted, but we still have to parse it because the value has to be a string since it's a text box.

client/src/components/SearchResultsOffset.js
davidsmejia: it basically refreshes the page

client/src/components/SearchResultsOffset.js
davidsmejia: Yeah we technically dont have to parse it i'm pretty sure it would work but here the enteredPageNumber should always be a string of numbers and the form should manage if/when you can submit it

client/src/components/SearchResultsOffset.js
davidsmejia: i rewrote this for clarity

client/src/components/SearchResult.js
davidsmejia: I think we would need the attachment in the material serializer in order to show the link to it?

client/src/components/SearchResultsFilters.js
```diff
@@ -0,0 +1,111 @@
...
+              checked={hasFacet('organism', organism.key)}
+              onChange={({ target: { checked } }) => {
+                toggleFacet(checked, 'organism', organism.key)
+                goToSearchResults(true)
+              }}
+            />
+          ))}
+        </Box>
+      )}
+      {facets.category && (
```
kurtwheeler, line 76: This should be `facets.publication_information` or something right?

client/src/theme/index.js
```diff
@@ -21,36 +21,46 @@ const theme = {
...
-        )}
-        `
+      extend: (props) =>
+        applyAll(
+          applyWhen(
+            props.open && !props.plain,
+            `background-color: ${normalizeColor('black-tint-80', props.theme)};`
+          ),
+          applyWhen(
+            props.plain && props.plain,
```
kurtwheeler, line 37:
```suggestion
            props.plain,
```

PR #160, opened 2020-05-06
adds imported and import_source to material document

davidsmejia:
## Issue Number

Materials search result was missing `imported` and `import_source` this adds them.


api/resources_portal/views/document_views.py
```diff
@@ -47,6 +47,8 @@ class MaterialDocumentSerializer(serializers.Serializer):
     organization = serializers.SerializerMethodField(read_only=True)
     mta_attachment = serializers.SerializerMethodField(read_only=True)
     additional_metadata = serializers.SerializerMethodField(read_only=True)
+    imported = serializers.BooleanField(read_only=True)
```
Philinphiladelphia, line 50: This should also be added to the class meta on line 66.

davidsmejia:
> Look good! Just one question: should import_source be a filterable field on the search endpoint?
@Philinphiladelphia i'm not sure, i don't really see anywhere we would need that on the front end yet.

api/resources_portal/views/document_views.py
davidsmejia: good catch!

Philinphiladelphia:
LGTM!

Issue #142, opened 2020-04-27
Set up notifications for invitations


kurtwheeler:
### Context

https://github.com/AlexsLemonade/resources-portal/issues/51
https://github.com/AlexsLemonade/resources-portal/issues/53

### Problem or idea

When a user receives an invitation they should be notified. When a user accepts another user's invitation, the inviter should be notified.

### Solution or next step

Trigger, track, and test notifications for the invitations table/route.

Issue #51, opened 2020-03-26
Create notification table


kurtwheeler:
### Context

I've created a set of end to end tests for us to implement: https://github.com/AlexsLemonade/resources-portal/issues/49

### Problem or idea

We've decided to go with a notification table! I'm leaving the rest of the ticket as it was to show why/what this should be, but now instead of a question ticket it's an implementation ticket!

I would like to be able to test what notifications go to who for each of these tests. I'd like it to be easy, and automatically fail the test if any unexpected notifications happen. This way we won't have to test that notifications don't happen in each test, we can just test what notifications should happen and then we'll be very explicit about them so we'll get good testing around our notification system.

### Solution or next step

So far I've come up with two ideas of how we can do this:
#### We could create a notification table tracking what notifications went to which users.
Pros:
  * This could potentially be useful if we ever wanted to analyze how many notifications we send to gauge how chatty our app is.
  * This could potentially be useful when looking into non-compliant users?

Cons:
  * It's an entire table we might not need
  * If we had 1000 users generate 100 notifications each it'd 100k rows, so we're looking at fairly tiny data, but it's still more data we're paying to have in our database.

#### OR

 We could just have a notify function that we use to trigger notifications and then mock that out in all the tests. Then we just have to compare the calls to that mock function to the expected notifications.

@dvenprasad @cgreene @arielsvn @davidsmejia I'd appreciate any thoughts or opinions on this.

cgreene:
I think I like the notification table. If it ends up being too large/costly, we can always trim it (or rotate it, or similar). 100k rows isn't that much, and that's at a level where we have far more users than we're expecting with the initial rollout. Plus easier debugging is always a win.

PR #161, opened 2020-05-06
Add invitations notifications

Philinphiladelphia:
## Issue Number

#51, #142 

This adds the notifications table and adds it to the invitations route. It also adds functional tests for this.

Two stories are addressed in this PR because it provides better visibility on how the design of notifications impacts developer usability.

## Types of changes

## Functional tests

Functional tests are found in api/resources_portal/test/views/test_organization_invitation.py

## Checklist

## Screenshots


api/resources_portal/models/notification.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+    associated_organization = models.ForeignKey(
+        Organization, blank=False, null=True, on_delete=models.CASCADE
+    )
+    associated_material = models.ForeignKey(
+        Material, blank=False, null=True, on_delete=models.CASCADE
+    )
+
+    @computed(models.TextField(null=False, blank=False))
+    def message(self):
```
Philinphiladelphia, line 60: This block is ugly as sin, but I couldn't figure out a better way to do it. Suggestions would be very welcome.

davidsmejia:
you will need to merge dev and change your migration name prefix with 007 and depend on `0006_auto_20200506_1330` also adding a migrations label so we can avoid conflicts like this in the future or plan accordingly

api/resources_portal/models/notification.py
davidsmejia: https://www.python.org/dev/peps/pep-3103/

api/resources_portal/models/notification.py
```diff
@@ -0,0 +1,92 @@
...
+        if self.notification_type == "ORG_REQUEST_CREATED":
+            return f"{self.associated_user.username} is requesting to to join {self.associated_organization.name}."
```
davidsmejia, line 62:
```suggestion
            return f"{self.associated_user.username} is requesting to join {self.associated_organization.name}."
```

api/resources_portal/models/notification.py
davidsmejia: or something like this 
```python

alert_messages = {
"ORG_REQUEST_CREATED" : "{} is requesting to to join {}.".format(self.associated_user.username, self.associated_organization.name)
...
}

try:
    message = alert_messages[self.response_type]
except ValueError:
    raise ValueError(f'"{self.notification_type}" is not a valid notification type')
```

api/resources_portal/models/notification.py
```diff
@@ -0,0 +1,92 @@
...
+        ("APPROVE_REQUESTS_PERM_GRANTED", "APPROVE_REQUESTS_PERM_GRANTED"),
+        ("TRANSFER_REQUESTED", "TRANSFER_REQUESTED"),
+    )
+
+    objects = models.Manager()
+
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    notification_type = models.CharField(max_length=32, choices=NOTIFICATION_TYPES)
+    notified_user = models.ForeignKey(
```
kurtwheeler, line 42: Can we also record what email we sent the notification to? This way if a user is complaining about their notifications not coming through and then changes their email we'll have a trail.

api/resources_portal/models/notification.py
kurtwheeler: I like the `alert_messages` dict idea!

api/resources_portal/test/views/test_organization_invitation.py
```diff
@@ -84,36 +89,126 @@ def test_request_approved_by_request_reciever_suceeds(self):
...
 
+        invitation = OrganizationInvitation.objects.get(pk=self.invitation.id)
+        requester = invitation.requester
+        organization = invitation.organization
+
         response = self.client.put(self.url, invitation_json)
+
+        # On acceptance, the invitation gets deleted
         self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertTrue(requester in organization.members.all())
```
kurtwheeler, line 100:
```suggestion
        self.assertTrue(invitation.requester in invitation.organization.members.all())
```

api/resources_portal/test/views/test_organization_invitation.py
```diff
@@ -84,36 +89,126 @@ def test_request_approved_by_request_reciever_suceeds(self):
...
+        invitation_json["status"] = "ACCEPTED"
 
         invitation = OrganizationInvitation.objects.get(pk=self.invitation.id)
         requester = invitation.requester
         organization = invitation.organization
 
-        self.assertEqual(invitation.status, "ACCEPTED")
+        response = self.client.put(self.url, invitation_json)
+
+        # On acceptance, the invitation gets deleted
```
kurtwheeler, line 120: It doesn't really get deleted does it? And also you aren't really testing for that here are you?

api/resources_portal/test/views/test_organization_invitation.py
```diff
@@ -84,36 +89,126 @@ def test_request_approved_by_request_reciever_suceeds(self):
...
+        invitation_json["status"] = "REJECTED"
+
+        invitation = OrganizationInvitation.objects.get(pk=self.invitation.id)
+        requester = invitation.requester
+        organization = invitation.organization
 
         response = self.client.put(self.url, invitation_json)
+
         self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertFalse(requester in organization.members.all())
```
kurtwheeler, line 164:
```suggestion
        self.assertFalse(invitation.requester in invitation.organization.members.all())
```
Same thought here/

api/resources_portal/test/views/test_organization_invitation.py
```diff
@@ -84,36 +89,126 @@ def test_request_approved_by_request_reciever_suceeds(self):
...
+        invitation_json["status"] = "INVALID"
+
+        invitation = OrganizationInvitation.objects.get(pk=self.invitation.id)
+        requester = invitation.requester
+        organization = invitation.organization
+
+        response = self.client.put(self.url, invitation_json)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertFalse(requester in organization.members.all())
```
kurtwheeler, line 206:
```suggestion
        self.assertFalse(invitation.requester in invitation.organization.members.all())
```
Same thought here.

api/resources_portal/views/organization_invitations.py
```diff
@@ -28,16 +28,28 @@ class OrganizationInvitationViewSet(viewsets.ModelViewSet):
...
+        if new_status == "ACCEPTED":
             invitation.organization.members.add(invitation.requester)
-            # TODO: Add notification of acceptance
-        elif old_status == "PENDING" and new_status == "REJECTED":
-            pass
-            # TODO: Add notification of rejection
-        elif old_status == "PENDING" and new_status == "INVALID":
-            pass
-            # TODO: Add notification of invalid request reciever
+            invitation.delete()
```
kurtwheeler, line 33: Don't delete the invitations! This data won't cost much to store and may one day be useful in debugging.

api/resources_portal/views/organization_invitations.py
```diff
@@ -28,16 +28,28 @@ class OrganizationInvitationViewSet(viewsets.ModelViewSet):
...
+
+        notification_type = f"ORG_{invitation.invite_or_request}_{new_status}"
+
+        notified_user = User
+        associated_user = User
```
kurtwheeler, line 38:
```suggestion
```
I don't believe these are necessary.

api/resources_portal/models/notification.py
Philinphiladelphia: Much better- thank you.

api/resources_portal/models/notification.py
Philinphiladelphia: Good catch!

api/resources_portal/models/notification.py
Philinphiladelphia: Good point.

api/resources_portal/models/notification.py
davidsmejia: I'm not sure if @kurtwheeler would go for it but i think this is slick
```python 
>>> class Test(object):
...     def __init__(self, a, b):
...         self.a = a
...         self.b = b
...
>>> a = "one"
>>> b = Test("two", "three")
>>> c = Test(a, b)
>>> demo = "{a} {b.b}"
>>> demo.format(**c.__dict__)
'one three'
```

davidsmejia: though either way you will want to determine your attributes to apply to the string before hand so you don't get an exception when creating the templates

api/resources_portal/views/organization_invitations.py
davidsmejia: we should be doing soft delete if we need things to be deleted or look deleted

api/resources_portal/models/notification.py
kurtwheeler: So I meant can we get an `email_address` field that gets set in `send_email_notification` with whatever email address was notified?

api/resources_portal/models/notification.py
```diff
@@ -0,0 +1,86 @@
...
+    associated_organization = models.ForeignKey(
+        Organization, blank=False, null=True, on_delete=models.CASCADE
+    )
+    associated_material = models.ForeignKey(
+        Material, blank=False, null=True, on_delete=models.CASCADE
+    )
+
+    @computed(models.EmailField(blank=False, null=False))
+    def email(self):
+        return self.notified_user.email
```
kurtwheeler, line 55: Oh I see, you added it here. Can we not do this as a computed field and instead set it when we send the notification so it cannot be changed by the user?

api/resources_portal/models/notification.py
```diff
@@ -0,0 +1,86 @@
...
+            "ORG_INVITE_REJECTED": lambda: f"{self.associated_user.username} has rejected your request to join {self.associated_organization.name}.",
+            "ORG_REQUEST_REJECTED": lambda: f"{self.associated_user.username} has rejected your invitation to join {self.associated_organization.name}.",
+            "ORG_INVITE_INVALID": lambda: f"You no longer have permissions to add members to {self.associated_organization.name}. Your pending invitations have been canceled.",
+            "ORG_REQUEST_INVALID": lambda: f"{self.associated_user.username} no longer has permissions to add members to {self.associated_organization.name}. Please resubmit your request.",
+            "MTA_UPLOADED": lambda: f"{self.associated_user.username} uploaded an MTA for {self.associated_material.title}.",
+            "APPROVE_REQUESTS_PERM_GRANTED": lambda: f"{self.associated_user.username} granted you permission to approve material transfer requests in {self.associated_organization.name}.",
+            "TRANSFER_REQUESTED": lambda: f'{self.associated_user.username} requested transfer of "{self.associated_material.title}".',
```
kurtwheeler, line 70: I don't understand why our linter isn't complaining about these lines being >100 characters... That's our project's limit. I know this block has been tricky to write well, but these lines are very very long.

I Think this might be a decent way to break these up: https://stackoverflow.com/a/54950733?

api/resources_portal/test/views/test_organization_invitation.py
```diff
@@ -84,36 +89,118 @@ def test_request_approved_by_request_reciever_suceeds(self):
...
+
+        invitation_json = self.client.get(self.url).json()
+
+        self.invitation.invite_or_request = "REQUEST"
+        self.invitation.save()
+        invitation_json["status"] = "INVALID"
 
         invitation = OrganizationInvitation.objects.get(pk=self.invitation.id)
         requester = invitation.requester
         organization = invitation.organization
```
kurtwheeler, line 174: Can we remove these two lines and just reference `invitation.*` where we're using them?

api/resources_portal/test/views/test_organization_invitation.py
```diff
@@ -84,36 +89,118 @@ def test_request_approved_by_request_reciever_suceeds(self):
...
+
+        invitation_json = self.client.get(self.url).json()
+
+        self.invitation.invite_or_request = "REQUEST"
+        self.invitation.save()
+        invitation_json["status"] = "REJECTED"
+
+        invitation = OrganizationInvitation.objects.get(pk=self.invitation.id)
+        requester = invitation.requester
+        organization = invitation.organization
```
kurtwheeler, line 134: Can we remove these two lines and just reference `invitation.*` where we're using them?

api/resources_portal/test/views/test_organization_invitation.py
```diff
@@ -84,36 +89,118 @@ def test_request_approved_by_request_reciever_suceeds(self):
...
+
+        invitation_json = self.client.get(self.url).json()
+
+        self.invitation.invite_or_request = "INVITE"
+        self.invitation.save()
+        invitation_json["status"] = "ACCEPTED"
 
         invitation = OrganizationInvitation.objects.get(pk=self.invitation.id)
         requester = invitation.requester
         organization = invitation.organization
```
kurtwheeler, line 113: Can we remove these two lines and just reference `invitation.*` where we're using them?

PR #163, opened 2020-05-06
search but with better data

davidsmejia:
## Issue Number

N/A

This mainly just handles the correct search facets as they will be returned from the API... probably will change a little to reduce code... also updates the fake data from a recent locally ran api search response. This also will change slightly as we build it out.

Next step would be to tie in a dynamic api endpoint and have the search work when run with `docker-compose up` but keep using the fake data on stage until a staging api is available


client/src/components/SearchResult.js
```diff
@@ -180,10 +180,11 @@ export const RequestRequirements = ({ resource }) => {
 
   const requirements = []
   const MTA = 'Material Transfer Agreement'
-  if (resource.needs_abstract) requirements.push('Abstract')
-  if (resource.needs_irb) requirements.push('IRB')
-  if (resource.needs_mta) requirements.push(MTA)
-  if (resource.needs_shipping_info) requirements.push('Shipping Information')
+  // if (resource.needs_abstract) requirements.push('Abstract')
+  // if (resource.needs_irb) requirements.push('IRB')
```
kurtwheeler, line 184: Any particular reason for these to hang around?

davidsmejia: I didn't see them in the response yet I didn't know if I am waiting for an attachment for them or a different attribute name

kurtwheeler: Oh, looks like they're just missing from the elasticsearch view. I just push an UNTESTED :scream: commit to address that. Can you test it? (I'm working on another branch with migrations right now.)

davidsmejia: yeah, ill update the testData too

kurtwheeler:
Why are you doing this instead of just hitting the search endpoint?

PR #177, opened 2020-05-13
User permissions and auth

Philinphiladelphia:
## Issue Number

#168 

This adds authentication and permissions to the users route.

## Types of changes

This adds permissions in the users viewset and the relevant tests in the test_user file.

## Functional tests

Using unit tests so far.

## Checklist


api/resources_portal/test/views/test_user.py
```diff
@@ -46,18 +48,58 @@ class TestUserDetailTestCase(APITestCase):
 
     def setUp(self):
         self.user = UserFactory()
+        self.second_user = UserFactory()
+        self.superuser = UserFactory(is_superuser=True)
```
kurtwheeler, line 52: Can we have the default be unauthenticated instead of superuser?

api/resources_portal/views/user.py
```diff
@@ -20,14 +19,29 @@ class Meta:
...
+    # The user endpoint does not support delete or list, so we don't have to worry about permissions for these
+    def get_permissions(self):
+        if self.action == "retrieve" or self.action == "update" or self.action == "partial_update":
+            permission_classes = [IsAuthenticated, IsUserOrAdmin]
+        elif self.action == "create":
+            permission_classes = []
+        else:
+            permission_classes = [IsAuthenticated]
+
+        return [permission() for permission in permission_classes]
```
kurtwheeler, line 44: I think that only admin users should be able to list all users. Can you add that and test it?

Philinphiladelphia: yup! added and tested.

api/resources_portal/test/views/test_user.py
```diff
@@ -18,12 +18,24 @@ class TestUserListTestCase(APITestCase):
...
         # Don't post created_at and updated_at.
         self.user_data.pop("created_at")
         self.user_data.pop("updated_at")
 
+    def test_list_request_from_admin_succeeds(self):
+        self.client.force_authenticate(user=self.superuser)
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_list_request_from_non_admin_forbidden(self):
```
kurtwheeler, line 34: Deepa said:

> We should let non-admin users to retrieve users. I can see a use case where people can request to be added to the organization too.

and I believe she was referring to list. Non-admin users need to be able to find other users for invitations and such.

api/resources_portal/test/views/test_user.py
```diff
@@ -46,18 +58,68 @@ class TestUserDetailTestCase(APITestCase):
...
 
     def test_get_request_returns_a_given_user(self):
+        self.client.force_authenticate(user=self.user)
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_get_request_from_wrong_user_forbidden(self):
+        self.client.force_authenticate(user=self.second_user)
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
```
kurtwheeler, line 74: This shouldn't be forbidden.

api/resources_portal/test/views/test_user.py
```diff
@@ -46,18 +58,68 @@ class TestUserDetailTestCase(APITestCase):
...
+
+    def test_delete_request_from_admin_succeeds(self):
+        self.client.force_authenticate(user=self.superuser)
+        response = self.client.delete(self.url)
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+
+    def test_delete_request_from_non_admin_forbidden(self):
+        self.client.force_authenticate(user=self.user)
+        response = self.client.delete(self.url)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
```
kurtwheeler, line 125: Users can't delete themselves?

api/resources_portal/views/user.py
```diff
@@ -52,21 +41,41 @@ class Meta:
...
+        if self.action == "list" or self.action == "destroy":
+            permission_classes = [IsAdminUser]
+        elif (
+            self.action == "retrieve" or self.action == "update" or self.action == "partial_update"
+        ):
+            permission_classes = [IsAuthenticated, IsUserOrAdmin]
+        elif self.action == "create":
+            permission_classes = [AllowAny]
+        else:
+            permission_classes = [IsAuthenticated]
```
kurtwheeler, line 79: Ok so this just reiterates the comments I left in the tests, but I think `retrieve` and `list` should be moved down to the else, `destroy` should be moved into the second `elif` (so they have IsUserOrAdmin), and then the first `if` can be removed.

Philinphiladelphia:
@kurtwheeler I think I've addressed your feedback. Have a look!

Issue #172, opened 2020-05-11
Create subroute for User to Grants relationship


kurtwheeler:
### Context

#167

### Problem or idea

Users can be associated with multiple grants. Using the /users/id/grants route should give you all the grants for that user. It's also how you create or destroy these relationships.

### Solution or next step

Implement the /user/id/grants subroute.

Make sure to set up the appropriate authentication/permissions for each operation and to test them well.

PR #180, opened 2020-05-14
Add user grants route

Philinphiladelphia:
## Issue Number

#172 

This adds the nested route for the grants to users relationship. It also adds tests for 100% coverage on this route.

## Types of changes

## Functional tests

Tests included.

## Checklist

### DIFF
--- a/api/resources_portal/views/grant_user.py
+++ b/api/resources_portal/views/grant_user.py
@@ -6,7 +6,7 @@ from rest_framework.response import Response
 from rest_framework_extensions.mixins import NestedViewSetMixin
 
 from resources_portal.models import Grant, GrantOrganizationAssociation, GrantUserAssociation, User
-from resources_portal.views.relation_serializers import MaterialRelationSerializer
+from resources_portal.views.relation_serializers import UserRelationSerializer
 
 
 class OwnsGrant(BasePermission):
@@ -28,7 +28,7 @@ class CanManageOrganizationUsers(BasePermission):
                 request.user.has_perm(
                     "add_members_and_manage_permissions", association.organization
                 )
-                and request.data["user"] in association.organization.members
+                and association.organization.members.filter(id=request.data["id"]).exists()
             ):
                 return True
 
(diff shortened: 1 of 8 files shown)


api/resources_portal/views/grant_user.py
```diff
@@ -0,0 +1,73 @@
...
+    """Allows users to add grant-user relationships only if they have permission to manage
+    users on an organization associated with the grant and the user to add is in that organization."""
+
+    def has_permission(self, request, view):
+        grant = Grant.objects.get(pk=view.kwargs["parent_lookup_grants"])
+
+        for association in GrantOrganizationAssociation.objects.filter(grant=grant):
+            if (
+                request.user.has_perm(
+                    "add_members_and_manage_permissions", association.organization
```
kurtwheeler, line 29: This permission is for users to add members to organizations and manage their permissions. This should not relate to grants at all.

I think grants is one area where we do want to force the PI to be involved. They are essentially only used for credit purposes, so generally most members of the organization won't need to worry about them, just the PI's who want to get credit for their grants sharing the resources they create.

So basically, only an owner of a grant should be allowed to modify it or add additional users to it.

Philinphiladelphia: Gotcha. Currently, that is the case- all operations on this endpoint require the user to be an owner of the grant. This permission is to make sure that owners can only add users from organizations that are:
A. Associated with the grant
B. Organizations that the owner has some kind of permissions on

To remedy this issue, should I maybe create a new permission on organizations that's like "can_add_users_to_grants" or something like that?

Philinphiladelphia: Alright, I think this has been addressed.

api/resources_portal/views/grant_user.py
```diff
@@ -0,0 +1,48 @@
...
+from rest_framework.permissions import BasePermission, IsAuthenticated
+from rest_framework.response import Response
+
+from rest_framework_extensions.mixins import NestedViewSetMixin
+
+from resources_portal.models import Grant, GrantUserAssociation, User
+from resources_portal.views.relation_serializers import UserRelationSerializer
+
+
+class OwnsGrant(BasePermission):
```
kurtwheeler, line 12: Can we make this OwnsGrantOrIsAdmin? Admins should be able to do everything the owner can.

Philinphiladelphia: Yep! just done, also added a test for it.

Issue #164, opened 2020-05-06
Setup CORS


davidsmejia:
### Context

I was running the client with `docker-compose up` but was unable to use the api served from `localhost:8000` we should get this working so we can dev against the testing data

### Problem or idea

Implement CORS with [django-cors-headers](https://github.com/adamchainz/django-cors-headers) 

### Solution or next step

* [x] Install [django-cors-headers](https://github.com/adamchainz/django-cors-headers) 
* [x] set env vars so when we run `docker-compose up` whitelist contains `localhost:7000`
* [x] set env vars so so staging contains TBD
* [x] set env vars so so production contains TBD
* [ ] also set up headers for `AUTH_TOKEN` for all envs

davidsmejia:
Just need to verify the auth headers for this.

PR #182, opened 2020-05-15
client + api cors

davidsmejia:
## Issue Number

#164

What a journey this has been. I have written a dockerfile for the client and also tried using nginx as a proxy for the api... but having the api use the client network seems to be the easiest way to make this work without requesting to modify the host machine's `/etc/hosts` file

This seems to be the cleanest way to have the client join the api localhost networking context.

Also a concern but really just rolled the client api out to make it work...
I haven't decided if the api should use the api nouns or the client nouns ie `api.search.materials.find()` vs `api.search.resources.find()`

## Types of changes

## Functional tests

Ran locally with `rportal up`

## Checklist

## Screenshots


client/src/api/index.js
```diff
@@ -0,0 +1,54 @@
...
+
+  try {
+    const response = await fetch(url, config)
+    return {
+      isOk: true,
+      status: response.status,
+      response: await response.json()
+    }
+  } catch (e) {
+    console.log(e)
```
kurtwheeler, line 38: You want this in here?

davidsmejia: yeah until sentry becomes a thing

client/src/api/index.js
```diff
@@ -0,0 +1,54 @@
+import fetch from 'isomorphic-unfetch'
```
kurtwheeler, line 1: Wow what a package name!

client/src/pages/resources/[id]/index.js
```diff
@@ -1,74 +1,81 @@
...
-      <Box pad="large">
-        <Heading level="5" margin={{ top: '0', bottom: 'large' }}>
-          <Link href="/resources/[id]" as={`/resources/${resource.id}`}>
-            <Anchor label={resource.title} />
-          </Link>
-        </Heading>
+const ResourceDetailsPage = ({ resource }) => {
+  if (!resource)
+    return (
+      <Text>This resource is not available. If this is an error... whoa</Text>
```
Comment (kurtwheeler) [inline client/src/pages/resources/[id]/index.js:13]: The error message makes me think we won't actually intend to display this to the user ever, but it looks like we will if the API fails to respond correctly when the resource's deets are requested. Should this just say "This resource is not currently available. Please try again soon."?

Comment (davidsmejia) [reply client/src/pages/resources/[id]/index.js]: I updated the message, you are right that this will never get show and actually we will need to set up an error boundary so we can capture unrecoverable errors too. This one would probably a 404 style page if someone deletes a resource.

Issue #175, opened 2020-05-11
Add Search Facets endpoint for search


davidsmejia:
### Context

Search facets are only included if they apply to the search results.. rerendering this leads to inconsistent UI. Would it be possible to always include all facets with 0 results or add a separate endpoint for search facets for each search endpoint.

davidsmejia:
closed with #190

PR #185, opened 2020-05-16
Searches With Post Filters

davidsmejia:
## Issue Number

#175 

Aside: we might want to combine `has_publication` and `has_pre_print` into a single facet `publication` and have those as terms that would enable the below fix to be applicable to those options (which are presented in the same filter section)

The client needs to be able to see a way to expand search results when a single search facet is applied to the query. 

When you search for something like `zebrafish` you would perform the following search

Query: `/search/materials/?search=zebrafish`
Response:
```javascript
{
    "facets": {
        "category": {
            "DATASET": 4,
            "MODEL_ORGANISM": 4,
            "CELL_LINE": 2,
            "OTHER": 2,
            "PLASMID": 2,
            "PROTOCOL": 2,
            "PDX": 1
        }
    },
    "results": [15 results]
}
```
----
Currently if you want to start narrowing down by categories and you start with `DATASET` you would do something like

Query: `/search/materials/?search=zebrafish&category=DATASET`
Response:
```javascript
{
    "facets": {
        "category": {
            "DATASET": 4
        }
    },
    "results": [4 results]
}
```
-----
Automatically changing the outputs would mean overriding the name and filtering of the search fields ie (category, organism etc). That is not ideal.

The client knows when it wants to be able to show additional options. So it can instead use a different filter parameter that corresponds to the desired results filter. The attribute in the query tells the API that it would also like the aggregates for that field not to be affected by this filter. However the results will still be filtered by the specified categories.

Query: `/search/materials/?search=zebrafish&category_pf=DATASET`
Response:
```javascript
{
    "facets": {
        "category": {
            "DATASET": 4,
            "MODEL_ORGANISM": 4,
            "CELL_LINE": 2,
            "OTHER": 2,
            "PLASMID": 2,
            "PROTOCOL": 2,
            "PDX": 1
        }
    },
    "results": [4 results]
}
```

## Types of changes

## Functional tests

This probably should get a test but I wanted to discuss approach first.

## Checklist

## Screenshots

Basic search query with 15 results

Screen Shot 2020-05-16 at 12 22 58 PM

With post filter applied the results are limited to Dataset but the aggregates contain other options that are filtered by the search term.

Screen Shot 2020-05-16 at 12 23 19 PM

### DIFF
--- a/api/resources_portal/views/document_views.py
+++ b/api/resources_portal/views/document_views.py
@@ -10,6 +10,7 @@ from django_elasticsearch_dsl_drf.filter_backends import (
     FacetedSearchFilterBackend,
     FilteringFilterBackend,
     OrderingFilterBackend,
+    PostFilterFilteringFilterBackend,
 )
 from django_elasticsearch_dsl_drf.pagination import LimitOffsetPagination as ESLimitOffsetPagination
 from django_elasticsearch_dsl_drf.viewsets import DocumentViewSet
@@ -172,7 +173,9 @@ class MaterialDocumentView(DocumentViewSet):
         DefaultOrderingFilterBackend,
         CompoundSearchFilterBackend,
         FacetedSearchFilterBackend,
+        PostFilterFilteringFilterBackend,
     ]
+
     lookup_field = "id"
 
     # Define search fields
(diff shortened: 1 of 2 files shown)


davidsmejia:
@kurtwheeler i updated the docs could you take a look at those please

Issue #170, opened 2020-05-11
Create subroute for Organizations to Materials relationship


kurtwheeler:
### Context

https://github.com/AlexsLemonade/resources-portal/pull/167

### Problem or idea

Organizations can have a lot of Materials. Using the `/organization/id/materials` route should give you all the materials for that organization.

I'm also thinking we should move the POST route from `/materials` to this sub-route. This way a material cannot be created that doesn't belong to an organization.

### Solution or next step

Implement the `/organization/id/materials` subroute and remove POST from `/materials`.

Make sure to set up the appropriate authentication/permissions for each operation and to test them well.

Issue #178, opened 2020-05-13
Add and manage permissions for lab/project members


dvenprasad:
### Context
Splitting #11 because it was dealing with 2 separate problems. 

### Relevant Use Cases:
**Four**
Applicable to: Sharers

I have some clinical data which I would like to make available through this portal. My PI insists on being able to review the requests with our collaborator. I would like requesters to submit an abstract of their work when they request for the data. I want to give my PI and collaborator access to the requests and the abstracts, so we can review them together.

**System Actions:**
1. Sharers must be able to have some requirements, i.e descriptions or abstracts for requesting certain resources which they can specify when they list a resource. 
2. Sharers must be able to grant restricted access to view requests to certain resources

**Twenty-four**
Applicable to: Sharers

I am a postdoc who has received a grant and used it to generate some cell lines. My PI’s lab has a lab manager who helps take care of prepping and shipping. Sometimes we get a lot of requests and emails are a cumbersome way to keep track of what and when things need to be shipped. I would like to add the lab manager to the project so tracking isn’t as difficult. 

**System Actions:**
1. Postdocs must be able to add people to their “projects”

### Problem or idea

Figure out:
- how users with grants will add project members to grants. 
- How and what permissions they can grant the members of a project. 

### Solution or next step

Whiteboard and prototype solution.

dvenprasad:
**Whiteboarding session:** 
https://projects.invisionapp.com/freehand/document/rnMLORvPs

**Summary of whiteboarding session:**
- Users can create "Teams", it can be members associated with a project or members associated with a lab. It will give users the flexibility and removes the burden on us to manage postdoc- PI relationship. 

- For MVP, only people who already possess accounts with the portal can be added to a team. Invitations will be sent to people who don't have accounts, but the owner of the team will need to try add them again after they have created accounts. 

- For MVP, permissions will be as follows: (Cannot be modified, but owners can transfer ownership) 
  - Only Owners can:
       -  Delete resource (within a certain time frame)
       - Add new members to the team
       - Change permissions for everyone ~except owner users and self~ 
       - Transfer ownership/ add a co-owner
       - Remove members from the team (each team must have at least one member though?)
  
  - Everyone on the Team can:
       - Add Resources
       - Edit Resource, even if they didn't add them 
       - Archive Resource
       - See requests and request materials submitted by the requesters
       - Approve/reject requests

dvenprasad:
**Low -fis**
[...]

PR #187, opened 2020-05-19
Add organizations materials subroute

Philinphiladelphia:
## Issue Number

#170 

Adds organizations users route. The post/put functionalities are handled through the materials route, the organizations-materials route is pretty much just for listing.

## Types of changes

## Functional tests

Unit tests

## Checklist


api/resources_portal/views/material.py
```diff
@@ -45,22 +45,38 @@ def get_serializer_class(self):
         return MaterialSerializer
 
     def get_permissions(self):
-        if self.action == "update" or self.action == "partial_update" or self.action == "destroy":
+        if (
+            self.action == "create"
+            or self.action == "update"
+            or self.action == "partial_update"
```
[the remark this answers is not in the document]
kurtwheeler: ....

This is the `/materials` route. Very nice. As we discussed though, we should use the permission model outlined in https://github.com/AlexsLemonade/resources-portal/issues/178#issuecomment-629246397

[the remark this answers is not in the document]
Philinphiladelphia: I added the permissions in this PR. Have a look!

api/resources_portal/models/organization.py
```diff
@@ -33,10 +45,29 @@ class Meta:
...
+
+
+def assign_member_perms(user, organization):
+    for permission in MEMBER_PERMISSIONS:
+        assign_perm(permission[0], user, organization)
+
+
+def remove_member_perms(user, organization):
+    for permission in MEMBER_PERMISSIONS:
+        remove_perm(permission[0], user, organization)
```
kurtwheeler, line 65: I love these functions, they're so nice and simple. However I think they'd be better as methods of the Organization class. Then they won't need the second parameter and they won't need to be imported separately.

api/resources_portal/test/views/test_organization_material.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    Tests /organizations/<id>/materials operations.
+    """
+
+    def setUp(self):
+        self.organization = OrganizationFactory()
+        self.material1 = MaterialFactory()
+        self.material2 = MaterialFactory()
+
+    def test_get_material_with_uuid_not_found(self):
```
kurtwheeler, line 22: It's NBD but it's not a UUID.

api/resources_portal/test/views/test_organization_material.py
```diff
@@ -0,0 +1,43 @@
...
+
+        self.organization.materials.add(self.material1)
+        self.organization.materials.add(self.material2)
+
+        url = reverse("organizations-materials-list", args=[self.organization.id])
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.json()["count"], 2)
+
+    def test_get_request_from_unauthenticated_fails(self):
```
kurtwheeler, line 39: I don't think this should be the case. Anyone can list materials by organization, it could potentially be helpful one day in material discovery.

api/resources_portal/models/organization.py
Philinphiladelphia: Smart. Just modified it to work this way.

api/resources_portal/test/views/test_organization_material.py
Philinphiladelphia: Changed to id.

api/resources_portal/test/views/test_organization_material.py
Philinphiladelphia: Gotcha. Removed this requirement.

PR #190, opened 2020-05-21
Add MaterialFacetedSearchFilterBackend

davidsmejia:
## Issue Number

#175 

Less elegant than just adding post filters. This PR instead just implements a  `MaterialFacetedSearchFilterBackend` which is a subclass of `FacetedSearchFilterBackend` that overrides the `aggregate` method and instead of ignoring the `post_filter_fields` it will apply them to the every aggregate except it's own.

In addition to the above it moves the appropriate `filter_fields` to the `post_filter_fileds` and removes the docs for the old post filter fields.

## Types of changes

## Functional tests

Tested in the browser.

## Checklist

## Screenshots

Screen Shot 2020-05-21 at 2 30 24 PM


davidsmejia:
This could probably be generalized and used in another project we know 😎

Issue #173, opened 2020-05-11
Create subroute for User to Organizations relationship


kurtwheeler:
### Context

#167

### Problem or idea

Users can be associated with multiple organizations. Using the `/users/id/organizations` route should give you all the organizations for that user. It's also one of the ways to destroy these relationships. Note that this is not the way you create these relationships though, because that goes through the invitations route.

### Solution or next step

Implement the `/user/id/organization` subroute.

Make sure to set up the appropriate authentication/permissions for each operation and to test them well.

PR #191, opened 2020-05-21
User-organizations route

Philinphiladelphia:
## Issue Number

#173 

This adds the user-organization route and associated tests.

## Types of changes

## Functional tests

Unit tests boys

## Checklist

### DIFF
--- a/api/resources_portal/views/user_organizations.py
+++ b/api/resources_portal/views/user_organizations.py
@@ -1,6 +1,6 @@
 from rest_framework import viewsets
 from rest_framework.exceptions import MethodNotAllowed, PermissionDenied, ValidationError
-from rest_framework.permissions import BasePermission, IsAuthenticated
+from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 
 from rest_framework_extensions.mixins import NestedViewSetMixin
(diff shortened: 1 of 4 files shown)


api/resources_portal/views/user_organizations.py
```diff
@@ -0,0 +1,41 @@
...
+
+from rest_framework_extensions.mixins import NestedViewSetMixin
+
+from resources_portal.models import Organization, OrganizationUserAssociation, User
+from resources_portal.views.relation_serializers import OrganizationRelationSerializer
+
+
+class UserOrganizationViewSet(NestedViewSetMixin, viewsets.ModelViewSet):
+    queryset = Organization.objects.all().order_by("-created_at")
+    permission_classes = [IsAuthenticated]
```
kurtwheeler, line 14: I think this is close, but I think it's an issue that any user can delete my organizations. I think we also need to check that the authenticated user is the user who's organizations are being modified. 

While thinking about this I checked out OrganizationMemberViewSet. It looks like it has a similar problem. It uses `BelongsToOrganization`, which seems appropriate for GET, but I think we need to add a `IsUserOrOwner` for the DELETE operation on that route. Would you mind including that in this PR as well?

kurtwheeler: Whoops, I see how that's implemented in both places already. Disregard that!

Issue #171, opened 2020-05-11
Create subroute for Organizations to Grants relationship


kurtwheeler:
### Context

#167

### Problem or idea

Organizations can have be associated with multiple grants. Using the /organization/id/grants route should give you all the grants for that organization. It's also how you create or destroy these relationships.

### Solution or next step

Implement the /organization/id/grants subroute.

Make sure to set up the appropriate authentication/permissions for each operation and to test them well.

PR #196, opened 2020-05-22
Add organization grants route

Philinphiladelphia:
## Issue Number

#171 

Added organizations-grants route and tests for it.

## Types of changes

## Functional tests

Screen Shot 2020-05-22 at 5 15 09 PM

## Checklist

### DIFF
--- a/api/resources_portal/test/views/test_organization_grant.py
+++ b/api/resources_portal/test/views/test_organization_grant.py
@@ -11,25 +11,6 @@ from resources_portal.test.factories import GrantFactory, UserFactory
 fake = Faker()
 
 
-def transform_grant_json(grant_json):
-    user_list = []
-    for user in grant_json["users"]:
-        user_list.append(user.id)
-    grant_json["users"] = user_list
-
-    organization_list = []
-    for org in grant_json["organizations"]:
-        organization_list.append(org.id)
-    grant_json["organizations"] = organization_list
-
-    material_list = []
-    for material in grant_json["materials"]:
-        material_list.append(material.id)
-    grant_json["materials"] = material_list
-
-    return grant_json
-
-
 class OrganizationGrantTestCase(APITestCase):
     """
     Tests /organizations/<id>/grants operations.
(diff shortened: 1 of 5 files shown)


api/resources_portal/views/organization_grant.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class OwnsGrantAndOrganization(BasePermission):
+    def has_permission(self, request, view):
+        if view.action == "create":
+            grant = Grant.objects.get(pk=request.data["id"])
+            organization = Organization.objects.get(pk=view.kwargs["parent_lookup_organizations"])
+        else:
+            grant = Grant.objects.get(pk=view.kwargs["pk"])
+            organization = Organization.objects.get(pk=view.kwargs["parent_lookup_organizations"])
```
kurtwheeler, line 26: This line is the same in both conditions, it can be moved down so we only change the way we retrieve the grant.

api/resources_portal/views/organization_grant.py
```diff
@@ -0,0 +1,68 @@
...
+
+        return request.user in grant.users.all() and request.user == organization.owner
```
kurtwheeler, line 28: If the user isn't the owner of the org, we don't need to query the grants table at all. Can you move the check for the owner above the condition where we retrieve the grant and return False if they aren't? Something like this:
```python
organization = Organization.objects.get(pk=view.kwargs["parent_lookup_organizations"])

if not request.user == organization.owner:
    return False

if view.action == "create":
            grant = Grant.objects.get(pk=request.data["id"])
else:
...
```

api/resources_portal/test/views/test_organization_grant.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def test_post_request_associates_a_grant(self):
+        self.client.force_authenticate(user=self.organization1.owner)
+
+        grant = GrantFactory()
+        grant.users.add(self.organization1.owner)
+        grant.save()
+
+        url = reverse("organizations-grants-list", args=[self.organization1.id])
+        grant_json = transform_grant_json(model_to_dict(grant))
```
kurtwheeler, line 94: Why is this call necessary?

Philinphiladelphia: Call was removed, replaced with GET as discussed.

api/resources_portal/views/organization_grant.py
Philinphiladelphia: Good catch

api/resources_portal/views/organization_grant.py
Philinphiladelphia: 👍

Issue #146, opened 2020-04-27
Implement CRUD routes for MaterialRequest


kurtwheeler:
### Context

The materials_request table needs routes so users can request materials.

### Problem or idea

Create the materials-request routes. Implement appropriate permissions.

### Solution or next step

Implement django routes for the MaterialRequest model.

PR #202, opened 2020-05-29
Add material request route

Philinphiladelphia:
## Issue Number

#146 

Adds routes and tests for material-request

## Types of changes

## Functional tests

Unit tests

## Checklist


api/resources_portal/migrations/0001_initial.py
```diff
@@ -734,6 +742,25 @@ class Migration(migrations.Migration):
...
+            name="groups",
+            field=models.ManyToManyField(
+                blank=True,
+                help_text="The groups this user belongs to. A user will get all permissions granted to each of their groups.",
+                related_name="user_set",
+                related_query_name="user",
+                to="auth.Group",
+                verbose_name="groups",
+            ),
+        ),
```
kurtwheeler, line 763: Hey so we're actually at the point where I think we should switch back to adding new migrations, because we have the staging server up and I don't really want to have to re-init its database :)

Philinphiladelphia: Gotcha. I just separated it out (:

api/resources_portal/migrations/0001_initial.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 2.2.12 on 2020-06-01 17:24
+# Generated by Django 2.2.12 on 2020-05-08 15:15
```
kurtwheeler, line 1: I'm confused, why is this file still changing?

api/resources_portal/test/views/__init__.py
```diff
@@ -0,0 +1,4 @@
+from resources_portal.test.views.test_material_request import (
+    TestMaterialRequestListTestCase,
+    TestSingleMaterialRequestTestCase,
+)
```
kurtwheeler, line 4: Why is this necessary?

api/resources_portal/views/material_request.py
```diff
@@ -0,0 +1,173 @@
...
+            "created_at",
+            "updated_at",
+        )
+        read_only_fields = ("id", "created_at", "updated_at")
+
+
+class MaterialRequestDetailSerializer(MaterialRequestSerializer):
+    assigned_to = UserSerializer()
+    requester = UserSerializer()
+    material = MaterialSerializer()
```
kurtwheeler, line 38: Can you actually use the *RelationSerializer for users and materials as well? The non-relation serializers have nested attributes, which we don't want. For example I'd expect a response from this endpoint to have a user object for who it's assigned to, but I don't need to see every org that user belongs to.

api/resources_portal/test/views/__init__.py
Philinphiladelphia: Ah- this was my mistake. I had it here so I could run these tests specifically instead of all of them. I'll remove.

kurtwheeler: Ohhhh. You couldn't run just them without it?

Philinphiladelphia: I don't think so. In order to run specific tests, you need to make the module which contains the tests visible to the run_tests.sh script. At some point I'll make a PR which adds all the tests to the __init__.py so the script can see them, but it's out of scope for this.

api/resources_portal/models/material_request.py
```diff
@@ -12,6 +12,13 @@ class Meta:
 
     objects = models.Manager()
 
+    STATUS_CHOICES = (
+        ("PENDING", "PENDING"),
+        ("APPROVED", "APPROVED"),
+        ("REJECTED", "REJECTED"),
+        ("INVALID", "INVALID"),
```
kurtwheeler, line 19: It just occurred to me that we should also have a `CANCELLED` state for if the requester cancels the request. This is different than `REJECTED` in that it was cancelled by the requester, rather than the sharer.

api/resources_portal/views/material_request.py
```diff
@@ -0,0 +1,173 @@
...
+                CanViewRequests,
+            ]
+        elif self.action == "retrieve":
+            permission_classes = [IsAuthenticated, CanViewRequestsOrIsRequester]
+        elif self.action == "update" or self.action == "partial-update":
+            permission_classes = [IsAuthenticated, CanApproveRequestsOrIsRequester]
+        elif self.action == "destroy":
+            permission_classes = [
+                IsAuthenticated,
+                IsRequester,
```
kurtwheeler, line 97: Hmm, so it looks like currently we're just allowing the user to delete rather than cancel their requests.

@davidsmejia do you think allowing these to be deleted is good or should we add a `CANCELLED` state that the requester can update the request to? I'm leaning towards having `CANCELLED` but I could maybe be convinced otherwise.

davidsmejia: I agree with your leanings. We should show cancelled for historical and reporting purposes

api/resources_portal/migrations/0002_auto_20200604_2003.py
```diff
@@ -0,0 +1,28 @@
...
+                    ("APPROVED", "APPROVED"),
+                    ("REJECTED", "REJECTED"),
+                    ("INVALID", "INVALID"),
+                    ("CANCELLED", "CANCELLED"),
+                ],
+                default="PENDING",
+                max_length=32,
+            ),
+        ),
+    ]
```
kurtwheeler, line 28: Wait, can this one be rolled up into the initial migration? (Sorry if I've been confusing about this. Basically once we have a staging server up I want to start adding additional migrations, but let's start that after this PR since I gotta re-init the staging database anyway.

api/resources_portal/views/material_request.py
```diff
@@ -0,0 +1,176 @@
...
+                material_request.requester_signed_mta_attachment = serializer.validated_data[
+                    "requester_signed_mta_attachment"
+                ]
+            if "status" in request.data and request.data["status"] != material_request.status:
+                if serializer.validated_data["status"] != "CANCELLED":
+                    return Response(status=403)
+                else:
+                    material_request.status = serializer.validated_data["status"]
+        else:
+            if "status" in request.data:
```
kurtwheeler, line 164: Can we also check that `request.data["status"] != "CANCELLED"`? I'd like for that state to only be set by the requester, otherwise the request should be `REJECTED`.

Philinphiladelphia: sure! done.

Issue #199, opened 2020-05-27
API requests (from me not the client)


davidsmejia:
### Context

I'm going to keep a list of API changes that I would like to see here. 

* [x] Material detail response (i think the serialization is out of date) needs the missing attributes
* [x] I think we need to remove the non nested contact user attributes from the model and also the search results
* [x] on material we should consider making the attribute `needs_mta` to be computed

PR #203, opened 2020-06-01
Update api

Philinphiladelphia:
## Issue Number

#199 

This updates the Material list and detail serializer to add missing attributes and remove redundant contact_email and contact_name attributes.

This PR now includes computed needs_mta.

## Functional tests

Tests pass with this modification

## Checklist


api/resources_portal/models/material.py
```diff
@@ -69,15 +69,17 @@ class Meta:
...
-    contact_name = models.TextField(blank=True, null=True)
-    contact_email = models.TextField(blank=True, null=True)
     publication_title = models.TextField(blank=True, null=True)
     pre_print_doi = models.TextField(blank=True, null=True)
     pre_print_title = models.TextField(blank=True, null=True)
     citation = models.TextField(blank=True, null=True)
     additional_info = models.TextField(blank=True, null=True)
     embargo_date = models.DateField(blank=True, null=True)
 
+    @computed(models.BooleanField(null=False))
```
davidsmejia, line 79: It is my understanding that this is required if we need the value in the DB but I don't think that that is required. I could be mistaken though. i think we just need it on instances of the model

api/resources_portal/migrations/0001_initial.py
```diff
@@ -734,6 +720,25 @@ class Migration(migrations.Migration):
...
+            name="groups",
+            field=models.ManyToManyField(
+                blank=True,
+                help_text="The groups this user belongs to. A user will get all permissions granted to each of their groups.",
+                related_name="user_set",
+                related_query_name="user",
+                to="auth.Group",
+                verbose_name="groups",
+            ),
+        ),
```
kurtwheeler, line 741: Same comment as the other PR, I think we should switch to adding rather than replacing migrations. Also note that you've got several PRs making migrations simultaneously so you'll need to merge and regenerate them one by one. (I know it's at least two, it might even be more though.)

Issue #181, opened 2020-05-14
Stop hard deleting objects.


kurtwheeler:
### Context

We probably want to keep records of what we've had in our system including users, organizations, materials, etc.

### Problem or idea

We should have a `is_deleted` flag on our all fields that we set to true when an object is deleted rather than actually deleting the record from our DB.

### Solution or next step

Set this up. I think Django has some kind of support for this, so it can probably be configured in some manner.

Note that Materials should be archivable, which might be a bit different than a soft-delete. I think we'll need an `is_archived` field as well.

kurtwheeler:
Change the base object manager for all the routes to ignore deleted objects.

PR #204, opened 2020-06-01
Add soft-delete

Philinphiladelphia:
## Issue Number

#181 

This adds soft deletion to all objects. It also adds tests.

## Types of changes

## Functional tests

Added tests to the views

## Checklist


kurtwheeler:
Looks like this branch has picked up some merge commits :(

api/resources_portal/models/user.py
```diff
@@ -1,18 +1,32 @@
...
+
+
+class DeletedObjectsManager(models.Manager):
+    def get_queryset(self):
+        return super().get_queryset().filter(deleted=True)
+
+
 class User(AbstractUser):
+    objects = NonDeletedObjectsManager()
+    deleted_objects = DeletedObjectsManager()
```
kurtwheeler, line 23: Why are you doing it like this instead of with `safedelete` like the other models?

api/resources_portal/test/views/test_grant.py
```diff
@@ -94,3 +94,10 @@ def test_cannot_update_someone_elses_grant(self):
         grant_json["title"] = "New Title"
         response = self.client.put(self.url, grant_json)
         self.assertEqual(response.status_code, 403)
+
+    def test_delete_only_soft_deletes_objects(self):
+        self.client.force_authenticate(user=self.grant.users.first())
+        grant_id = self.grant.id
+        response = self.client.delete(self.url)
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+        self.assertEqual(Grant.deleted_objects.filter(id=grant_id).count(), 1)
```
kurtwheeler, line 103: Can you also test that subsequent GET requests return a 404?

api/resources_portal/test/views/test_organization_invitation.py
```diff
@@ -213,3 +213,27 @@ def test_put_request_without_authentication_fails(self):
...
+        self.client.force_authenticate(user=None)
+        response = self.client.delete(self.url)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_delete_request_only_soft_deletes_objects(self):
+        self.client.force_authenticate(user=self.invitation.requester)
+        invitation_id = self.invitation.id
+        response = self.client.delete(self.url)
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+        self.assertEqual(OrganizationInvitation.deleted_objects.filter(id=invitation_id).count(), 1)
```
kurtwheeler, line 239: This seems identical to `test_delete_request_deletes_invitation` except with a different final assertion. I think they could just be combined to a single test with both assertions. (This is actually basically what I was requesting with the subsequent GET requests.)

api/resources_portal/models/user.py
Philinphiladelphia: safe-delete doesn't play nice with the AbstractUser parent class. I looked up the code for AbstractUser on GitHub, and tried quite a few ways of inheriting/combining AbstractUser and SafeDeleteModel, but nothing seemed to work. Anything that directly overrides the "objects" property of the model causes failure. I also tried some solutions where I combined the UserManager and SafeDeleteManager, but no bones on that either.

Philinphiladelphia: Is it important enough that I should give it some more effort?

api/resources_portal/views/user.py
```diff
@@ -60,6 +61,14 @@ def create(self, request, *args, **kwargs):
 
         return response
 
+    def destroy(self, request, *args, **kwargs):
+        user = User.objects.get(pk=kwargs["pk"])
+        self.check_object_permissions(request, user)
+        user.deleted = True
+        user.save()
+
+        return Response(status=204)
```
kurtwheeler, line 70: I think this logic should be done on the model instead of the view. However ideally we could use the `softdelete` package here so the behavior can be consistent.

Issue #194, opened 2020-05-22
Attachments routes


kurtwheeler:
### Context

We have an Attachment model.

https://github.com/AlexsLemonade/resources-portal/issues/112

### Problem or idea

We need routes for these. They will be uploaded before their material is created, then updated with the material's ID.

### Solution or next step

Implement the attachment route.

PR #211, opened 2020-06-08
Add attachment route

Philinphiladelphia:
## Issue Number

#194 

Adds attachment routes and the relevant tests

## Types of changes

## Functional tests

Unit tests

## Checklist


api/resources_portal/models/attachment.py
```diff
@@ -15,17 +24,18 @@ class Meta:
...
 
     sequence_map_for = models.ForeignKey(
-        "resources_portal.Material",
-        blank=False,
+        "Material",
+        blank=True,
         null=True,
         on_delete=models.SET_NULL,
         related_name="sequence_maps",
-        help_text="The cell line this seq_map is for. Only valid for seq_map attachments.",
```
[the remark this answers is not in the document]
Philinphiladelphia: Nope, just got removed by accident as I was working with the field. Adding it back in.

api/resources_portal/views/attachment.py
```diff
@@ -0,0 +1,78 @@
...
+            permission_classes = [IsAuthenticated]
+        else:
+            permission_classes = [IsAuthenticated, HasModifyAttachmentOrIsAdminUser]
+
+        return [permission() for permission in permission_classes]
+
+    def create(self, request, *args, **kwargs):
+        if not (
+            MaterialRequest.objects.filter(requester=request.user, is_active=True).exists()
+            or MaterialRequest.objects.filter(assigned_to=request.user, is_active=True).exists()
```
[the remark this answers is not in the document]
Philinphiladelphia: This is quite a difficult thing to do because attachments aren't associated with an organization initially. The permissions to link an attachment to material request are already as you describe: anyone with the ability to approve requests can currently link attachments to the material request. The only permission I could maybe put here would be that the user has to be in any organization with an active material request, but that's pretty permissive as it is. Is that worth doing, or should we just let anyone upload an attachment? Or should we link attachments to an organization initially?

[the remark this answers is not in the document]
kurtwheeler: > The only permission I could maybe put here would be that the user has to be in any organization with an active material request, but that's pretty permissive as it is.

Yeah it's pretty permissive, but I don't think that's an issue. I think it's better than forcing the assignee to be the one to upload things.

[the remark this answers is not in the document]
Philinphiladelphia: Ok, I added this in. My code block for this is pretty ugly- if there's a way to simplify it, lmk. If not, there it is.

api/resources_portal/test/views/test_attachment.py
```diff
@@ -0,0 +1,161 @@
...
+    def test_get_request_from_unauthenticated_fails(self):
+        self.client.force_authenticate(user=None)
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_put_request_updates_a_attachment(self):
+        self.client.force_authenticate(user=self.user)
+        attachment_json = self.client.get(self.url).json()
+
+        attachment_json["sequence_map_for"] = attachment_json["sequence_map_for"]["id"]
```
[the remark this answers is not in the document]
Philinphiladelphia: I had a look at the API, and it turns out that it's not currently consistent anyway. Material and Organization do this different ways. I changed this one to take objects, but I should maybe do another PR to address that.

api/resources_portal/test/factories.py
```diff
@@ -109,6 +109,7 @@ class Meta:
     description = "A file for testing"
     s3_bucket = "https://bucket-name.s3.region.amazonaws.com/keyname"
     s3_key = "s3 key"
+    sequence_map_for = factory.SubFactory(MaterialFactory)
```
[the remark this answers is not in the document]
Philinphiladelphia: Done.

api/resources_portal/views/material_request.py
```diff
@@ -150,12 +152,15 @@ def update(self, request, *args, **kwargs):
 
         if request.user == material_request.requester:
             if "irb_attachment" in request.data:
-                material_request.irb_attachment = serializer.validated_data["irb_attachment"]
+                irb = serializer.validated_data["irb_attachment"]
+                material_request.irb_attachment = irb
+                # allow the assignee read access to the attachment
+                assign_perm("view_attachment", material_request.assigned_to, irb)
```
kurtwheeler, line 158: Limiting this to just the assignee seems overly restrictive I think. I'm not actually sure though so I'm gonna loop in @dvenprasad: Are IRBs so private that it'd be a problem to let anyone who can manage requests see them?

api/resources_portal/views/material_request.py
```diff
@@ -150,12 +152,15 @@ def update(self, request, *args, **kwargs):
...
 
             if "requester_signed_mta_attachment" in request.data:
-                material_request.requester_signed_mta_attachment = serializer.validated_data[
-                    "requester_signed_mta_attachment"
-                ]
+                signed_mta = serializer.validated_data["requester_signed_mta_attachment"]
+                material_request.requester_signed_mta_attachment = signed_mta
+                assign_perm("view_attachment", material_request.assigned_to, signed_mta)
```
kurtwheeler, line 163: Same answer to https://github.com/AlexsLemonade/resources-portal/pull/211/files#r442897370 applies here as well.

api/resources_portal/views/material_request.py
```diff
@@ -170,9 +175,9 @@ def update(self, request, *args, **kwargs):
                 material_request.status = serializer.validated_data["status"]
 
             if "executed_mta_attachment" in request.data:
-                material_request.executed_mta_attachment = serializer.validated_data[
-                    "executed_mta_attachment"
-                ]
+                executed_mta = serializer.validated_data["executed_mta_attachment"]
+                material_request.executed_mta_attachment = executed_mta
+                assign_perm("view_attachment", material_request.assigned_to, executed_mta)
```
kurtwheeler, line 180: Same answer to https://github.com/AlexsLemonade/resources-portal/pull/211/files#r442897370 applies here as well.

api/resources_portal/views/material_request.py
kurtwheeler: I think one potential solution to this could be to make `view_attachments` a permission for organizations, then once we make those granular enough owners could restrict which members can see attachments if they care.

api/resources_portal/views/attachment.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class HasViewAttachmentOrIsAdminUser(BasePermission):
+    def has_object_permission(self, request, view, obj):
+        return request.user.has_perm("view_attachment", obj) or request.user.is_staff
+
+
+class HasModifyAttachmentOrIsAdminUser(BasePermission):
+    def has_object_permission(self, request, view, obj):
+        return request.user.has_perm("modify_attachment", obj) or request.user.is_staff
```
kurtwheeler, line 39: Yeah when I see this here, I think we want these to be organization level, not request level, but I await @dvenprasad's confirmation.

api/resources_portal/views/material_request.py
kurtwheeler: Ok so Deepa and I talked. My last suggestion was even too granular. If someone can review a request for an organization, they should be able to view its attachments. Once again, `assigned_to` is just to know where the buck stops, not for permissions.

api/resources_portal/views/attachment.py
dvenprasad: Yes, we want the permissions on an organization level

api/resources_portal/views/attachment.py
```diff
@@ -25,28 +25,35 @@ class Meta:
         read_only_fields = ("id", "created_at", "updated_at")
 
 
+def user_has_perm_on_active_material_request(user, perm):
+    for organization in user.organizations.all():
+        if user.has_perm(perm, organization):
+            for material in organization.materials.all():
+                for request in material.requests.all():
+                    if request.is_active:
+                        return True
```
kurtwheeler, line 32: I'm concerned this could become a lot of queries at some point. They say premature optimization is the devil, so I'm not sure we should rewrite all of this to be a single query just yet. However perhaps we could use this https://django-guardian.readthedocs.io/en/stable/userguide/performance.html#prefetching-permissions to at least prevent a separate query for each org.

Additionally we should probably prefetch requests, since we know we want to loop over all of them. (I'll make a suggestion for that.)

api/resources_portal/views/attachment.py
```diff
@@ -25,28 +25,35 @@ class Meta:
         read_only_fields = ("id", "created_at", "updated_at")
 
 
+def user_has_perm_on_active_material_request(user, perm):
+    for organization in user.organizations.all():
+        if user.has_perm(perm, organization):
+            for material in organization.materials.all():
```
kurtwheeler, line 29:
```suggestion
            for material in organization.materials.all().prefetch_related('requests'):
```

api/resources_portal/views/attachment.py
```diff
@@ -25,28 +25,35 @@ class Meta:
...
+                        return True
+    return False
+
+
 class AttachmentDetailSerializer(AttachmentSerializer):
     sequence_map_for = MaterialRelationSerializer(many=False, read_only=True)
 
 
-class HasViewAttachmentOrIsAdminUser(BasePermission):
+class CanViewRequestsOrIsAdminUser(BasePermission):
```
kurtwheeler, line 40: Shouldn't the requester also be able to view the attachments for a material?

api/resources_portal/views/attachment.py
```diff
@@ -24,9 +26,16 @@ class Meta:
...
-    for organization in user.organizations.all():
+
+    # Retrieve all organization permissions in a single query
+    checker = ObjectPermissionChecker(user)
+    organizations = Organization.objects.all()
+    checker.prefetch_perms(organizations)
+
+    # Uses prefetch_related to retrieve all related objects in a single query
+    for organization in user.organizations.all().prefetch_related("materials"):
         if user.has_perm(perm, organization):
```
kurtwheeler, line 37: I think you need to use the `checker` here.

Philinphiladelphia: Ahh, you're right. Good catch.

Issue #208, opened 2020-06-08
ORCID / OAuth Backend Ticket


cgreene:
### Context

We are supporting login with ORCID.

### Problem or idea

We need to make sure the database has the components that we need to support login, and we also need to make sure the appropriate API endpoints are exposed.

### Solution or next step

We should implement this. It'd also be a good place to get a bit of test infrastructure built out and tested.

PR #212, opened 2020-06-15
Implement oauth

Philinphiladelphia:
## Issue Number

#208 

This adds a auth route for ORCID Oauth. It also removes the POST route for user, as that route doesn't really make sense now that users require an ORCID, access token and refresh token.

To begin the sign-in process, the user will click a button on the site which has a URL of this form:

```
https://orcid.org/oauth/authorize?
   client_id=APP-2AHZAK2XCFGHRJFM&
   response_type=code&
   scope=/authenticate&
   redirect_uri=http://localhost:8000/v1/search/
```

Where the redirect_uri is replaced with whatever page the button is on.

The redirect URI can also accept parameters, specifically multiple grant_ids and a single email. They should be in the below form:

`http://localhost:8000/v1/search/?grant_id=[GRANT_ID]&grant_id=[GRANT_ID]&email=[EMAIL]`

Note that for staging and local development the endpoint is set to `https://sandbox.orcid.org/oauth/authorize`

## Types of changes

## Functional tests

End-to-end tests are in a separate ticket.

## Checklist


api/resources_portal/middleware/oauth.py
```diff
@@ -0,0 +1,97 @@
...
+import furl
+import orcid
+import requests
+
+from resources_portal.models.grant import Grant
+from resources_portal.models.organization import Organization
+from resources_portal.models.user import User
+
+CLIENT_ID = "APP-2AHZAK2XCFGHRJFM"
+CLIENT_SECRET = "1aaa0a94-5605-4cd8-95a4-6724c2f7fd42"
```
Philinphiladelphia, line 12: Oh dear, I didn't realize this repo was public. David recommended that we use this compromised client secret for testing, then ask for a refresh once the infrastructure for storing it (like git crypt) is set up.

kurtwheeler: We already have the infrastructure for storing secrets: https://github.com/AlexsLemonade/resources-portal/settings/secrets

Basically what you want to do is read it in as an environment variable, then when the system is deployed we'll provide the real value from Github's secrets.

The one question that leaves is how it will work in development though?

Also just to note: it is definitely not best practices to include secrets in plaintext even if it were a private repo.

api/requirements.txt
```diff
@@ -47,3 +47,7 @@ django-elasticsearch-dsl==6.4.2
 elasticsearch-dsl==6.4.0
 elasticsearch==6.4.0
 drf_yasg
+
+# Orcid
+orcid
```
kurtwheeler, line 52: Can you set a version for this?

api/resources_portal/middleware/oauth.py
```diff
@@ -0,0 +1,97 @@
...
+def remove_code_parameter_from_uri(url):
+    url_obj = furl.furl(url)
+    url_obj.remove(["code"])
+    return url_obj.url
+
+
+class OAuthMiddleWare:
+
+    """ Intercepts requests containing a "code" parameter provided by ORCID OAuth.
+    The authorization code is checked against ORCID and will provide a access token."""
```
kurtwheeler, line 28:
```suggestion
    The authorization code is checked against ORCID and will provide an access token."""
```

api/resources_portal/middleware/oauth.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if not ("code" in request.GET.keys()):
```
kurtwheeler, line 34:
```suggestion
        if "code" not in request.GET.keys():
```

api/resources_portal/middleware/oauth.py
```diff
@@ -0,0 +1,97 @@
...
+            data = {
+                "client_id": CLIENT_ID,
+                "client_secret": CLIENT_SECRET,
+                "grant_type": "authorization_code",
+                "code": authorization_code,
+                "redirect_uri": uri,
+            }
+
+            # get user orcid info
+            URL = "https://sandbox.orcid.org/oauth/token"
```
kurtwheeler, line 51: We're probably also going to want to make this an env var so we can hit the not-sandbox URL in prod.

api/resources_portal/views/user.py
```diff
@@ -50,27 +28,11 @@ class UserViewSet(viewsets.ModelViewSet):
...
-            return CreateUserSerializer
-
-        return UserSerializer
+    http_method_names = ["get", "delete", "put", "patch", "head", "options"]
 
     def get_permissions(self):
         if self.action == "update" or self.action == "partial_update" or self.action == "destroy":
             permission_classes = [IsAuthenticated, IsUserOrAdmin]
-        elif self.action == "create":
-            permission_classes = [AllowAny]
```
kurtwheeler, line 73: If POSTs aren't allowed, how are users created?

Philinphiladelphia: Users are created through the OAuth route. If we want users to be created any other way, we need to get their ORCID, which can only be acquired if you have our client secret. We can still create routes which do this, but just POSTing to user doesn't make very much sense anymore to me. We can discuss further on a call if you like.

api/resources_portal/middleware/oauth.py
```diff
@@ -0,0 +1,96 @@
...
+            response = requests.post(URL, data=data, headers={"accept": "application/json"})
+            response_json = response.json()
+
+            user_queryset = User.objects.filter(orcid=response_json["orcid"])
+
+            # Create user if neccessary
+            if user_queryset.exists():
+                user = user_queryset.first()
```
kurtwheeler, line 59: `first()` returns None if nothing matches the query, so you could instead do something like:
```
user = User.objects.filter(orcid=response_json["orcid"]).first()

            # Create user if neccessary
            if not user:
```

It's a minor improvement, but I think its a bit cleaner.

api/resources_portal/middleware/oauth.py
Philinphiladelphia: Cool- done.

api/resources_portal/middleware/oauth.py
```diff
@@ -48,7 +49,11 @@ def __call__(self, request):
             }
 
             # get user orcid info
-            URL = "https://sandbox.orcid.org/oauth/token"
+            if OAUTH_ENVIRONMENT == "SANDBOX":
+                URL = "https://sandbox.orcid.org/oauth/token"
+            else:
+                URL = "https://orcid.org/oauth/token"
```
kurtwheeler: Why do this rather than just make the URL the env var?

Philinphiladelphia: Oh dear, this was from an older version- looks like it didn't save correctly. Fixing.

api/resources_portal/config/common.py
```diff
@@ -43,13 +43,19 @@ class Common(Configuration):
...
 
     ALLOWED_HOSTS = ["*"]
     ROOT_URLCONF = "resources_portal.urls"
     SECRET_KEY = os.getenv("DJANGO_SECRET_KEY")
     WSGI_APPLICATION = "resources_portal.wsgi.application"
 
+    # OAuth
+    CLIENT_SECRET = os.getenv("OAUTH_CLIENT_SECRET")
+    CLIENT_ID = "APP-2AHZAK2XCFGHRJFM"
+    OAUTH_URL = os.getenv("OAUTH_ENIVRONMENT", "https://sandbox.orcid.org/oauth/token")
```
kurtwheeler, line 57: This is defined in local and production, so this should get overwritten anyway. It's also confusing that it's loading in `OAUTH_ENVIRONMENT` for `OAUTH_URL`.

Philinphiladelphia: My bad. So I can just remove this line entirely?

kurtwheeler: I think so!

Philinphiladelphia: Sweet, it works. Good catch!

api/resources_portal/test/middleware/test_oauth.py
```diff
@@ -0,0 +1,42 @@
...
+        user = User.objects.get(pk=self.client.session["_auth_user_id"])
+
+        self.assertEqual(user.orcid, ORCID_AUTHORIZATION_DICT["orcid"])
+
+        self.assertEqual(user.email, MOCK_EMAIL)
+
+        self.assertTrue(self.grant1 in user.grants.all())
+        self.assertTrue(self.grant2 in user.grants.all())
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
kurtwheeler, line 42: Hmm, it just occurred to me that we should also test that login works after a user has already been created.

Issue #174, opened 2020-05-11
Stand up staging server and configure HTTPS for it


kurtwheeler:
### Context

The frontend has its first pages, including search. It'd be cool to be able to show it to people with data from the server.

### Problem or idea

There currently isn't any server to hook it up to.

### Solution or next step

Create and configure a staging stack so that we can easily show off demos with live (albeit fake) data.

PR #214, opened 2020-06-19
Deploy staging server continuously with github actions

kurtwheeler:
## Issue Number

#174 

This is most of the way there. The one issue I cannot figure out is how to refresh the `http_auth` for the elasticsearch client. Once it expires our search breaks. As far as I can tell, the elasticsearch libraries only accept a string for that setting and don't provide a way to refresh it, so I've open an issue with them about it: https://github.com/elastic/elasticsearch-dsl-py/issues/1377.

## Types of changes

## Functional tests

https://github.com/AlexsLemonade/resources-portal/runs/786004899?check_suite_focus=true worked!!!

## Checklist


kurtwheeler:
Could this be related? https://github.com/barseghyanartur/django-elasticsearch-dsl-drf/issues/108

infrastructure/database.tf
```diff
@@ -24,25 +24,25 @@ resource "aws_db_instance" "postgres_db" {
...
   engine_version = "9.6.11"
   auto_minor_version_upgrade = false
-  instance_class = "db.${var.database_instance_type}"
+  instance_class = var.database_instance_type
   name = "resources_portal"
-  port = "${var.database_port}"
-  username = "${var.database_user}"
-  password = "${var.database_password}"
+  port = "5432"
+  username = "rppostgresuser"
```
davidsmejia, line 30: 🤔

davidsmejia:
Still looking this over but on stage we arent serving static files.. i see media in urls.py but no mention of static

Also the link in config/common.py should probably be https://docs.djangoproject.com/en/3.0/howto/static-files/deployment/

infrastructure/database.tf
kurtwheeler: ?

kurtwheeler:
Ok @davidsmejia I believe I am done with this PR!

kurtwheeler:
Ok I've added those env vars to github actions!

kurtwheeler:
Hmm, you know, I'm not entirely 100% sure. We might not need subnets?

Issue #57, opened 2020-03-26
EndToEndTest New resource is listed and requested


kurtwheeler:
### Context

I've created a set of end to end tests for us to implement: https://github.com/AlexsLemonade/resources-portal/issues/49

This issue relies on the test data created by https://github.com/AlexsLemonade/resources-portal/issues/50

### This test should test:

A new resource is listed for PrimaryLab and a request is fulfilled for it. (tests that Postdoc has permission for new resources)

#### The test should do that by making the following API calls:
1. The PrimaryProf lists a new resource with PrimaryLab.
2. SecondaryProf requests the resource.
3. Postdoc approves the request.
4. SecondaryProf upload the signed MTA.
5. Postdoc uploads the executed MTA.

#### This test should test that the following notifications occur and no more (see https://github.com/AlexsLemonade/resources-portal/issues/51):
- [ ] The Postdoc receives a notification that a resource was requested
- [ ] SecondaryProf is notified that her request was approved.
- [ ] Postdoc receives notification that SecondaryProf has signed MTA.

dvenprasad:
Step 3 should be just Postdoc approves request
4. SecondaryProf is notified to sign and upload the signed MTA.
5. Postdoc is notified and signed and uploads the executed MTA. 

So, for notifications, I would add
- [ ] Secondary Prof is prompted to sign and upload MTA. ( this can be in the same notification as approval notification, so feel free to ignore this one.)
- [ ] Postdoc receives notification that SecondaryProf has signed MTA.

IRB is only for review purposes, no action needs to be taken on it by Sharers.

PR #224, opened 2020-06-25
New resource listed and requested end to end test

Philinphiladelphia:
## Issue Number

#57 

This adds an end to end test which goes through the flow of a material being listed, requested, and transferred.

Other changes to help with this flow include:
- Addition of a few notifications to the `material-request` route
- Removal of `organization_user_setting` from the dev_data, as it's created automatically and it was resulting in duplicates.
- Refactoring of `populate_test_database` command so that I can access it as a function
- Modification of field `assigned_to` on material_request to be computed. Previously it was assigned on the route which didn't make much sense.

Question for Kurt: 
Should I add a "end_to_end.md" file which describes what each of the tests does?

## Types of changes

## Functional tests

I don't know if this section applies. Otherwise it would be
Screen Shot 2020-06-25 at 12 22 47 PM

## Checklist


api/resources_portal/models/material_request.py
```diff
@@ -62,3 +59,11 @@ class Meta:
...
 
     status = models.CharField(max_length=32, choices=STATUS_CHOICES, default="PENDING")
+
+    @computed(
+        models.ForeignKey(
+            User, blank=False, null=False, on_delete=models.CASCADE, related_name="assignments"
+        )
+    )
+    def assigned_to(self):
+        return self.material.contact_user
```
kurtwheeler, line 69: I think this should be a field of its own that is set to this by default, but at some point we'll want to let people change it. (For example, someone could be the contact_user for a material, but then hand out fulfillments on a rotating basis or something to share the workload.)

kurtwheeler:
> Should I add a "end_to_end.md" file which describes what each of the tests does?

I think the description in the docs of each like you're doing is sufficient!

Issue #139, opened 2020-04-24
Refactor Frontend Components for Resources


davidsmejia:
### Context

I want to refactor the resources so that we can dynamically generate the details and forms and which will help keep metadata fields consistent.

### Problem or idea

Pull out resources components to be generic and have them render the resources model.

### Solution or next step

![image](https://user-images.githubusercontent.com/1075609/80237107-296d5180-862a-11ea-98a4-37370bae579c.png)

PR #235, opened 2020-07-06
Create schemas for resources/materials

davidsmejia:
## Issue Number

#218 
#139 

Refactor the material and add the schema for material

I accidentally added some API clean up in this PR

I would recommend reviewing in the following order:
1. schemas/material.js
1. material configs in `components/resources/configs/`
1. the components that consume these changes
1. api changes
1. the remaining files like package.json and yarn.lock etc


client/src/helpers/getResourceValue.js
```diff
@@ -0,0 +1,8 @@
+const customResourceValues = {}
+
+export const getResourceValue = (resource, token) => {
+  if (token in customResourceValues) {
+    return customResourceValues[token](resource)
+  }
+  return resource[token] || resource.additional_metadata[token]
+}
```
kurtwheeler, line 8: What's this function do? As far as I can tell `customResourceValues` will always be empty, so it'll never return a match. Could you add a comment explaining?

client/src/components/resources/index.js
```diff
@@ -1,45 +1,9 @@
...
-export const ResourceDetails = ResourceComponentGetter('ResourceDetails')
+export const resourceCategories = [
+  'CELL_LINE',
+  'PLASMID',
+  'PROTOCOL',
+  'DATASET',
+  'MODEL_ORGANISM',
+  'PDX',
+  'OTHER'
+]
```
kurtwheeler, line 9: This seems to be repeating the list in `materials.js`. Is there an easy way to get them to use the same one so if it needs to be updated it only needs to happen in one place?

client/src/components/SearchInput.js
```diff
@@ -26,7 +28,11 @@ export default function SearchInput({ onChange, size = 'medium' }) {
   }
 
   // this is just the help make the select look more like the designs
-  const selectWidth = resourceType === 'ALL' ? '60px' : '156px'
```
kurtwheeler, line 29: Should the comment above this have been moved as well?

client/src/helpers/getResourceValue.js
davidsmejia: Yeah there needs to be some clarification on what the custom fallbacks will be but this is how that will work

client/src/components/resources/index.js
davidsmejia: yeah we can source the same list from the schema

client/src/components/SearchInput.js
davidsmejia: yup 👍

kurtwheeler:
BTW ESLint is unhappy with you.

client/src/components/resources/index.js
kurtwheeler: Ohhh, this is for routing isn't it? That explains why you're being so explicit.

client/src/components/resources/configs/cellLine.js
```diff
@@ -0,0 +1,65 @@
...
+      'cryopreservation',
+      'culture_conditions'
+    ]
+  },
+  {
+    quality: ['str_profile', 'passage_number', 'description']
+  }
+]
+
+export const form = [
```
kurtwheeler, line 32: You know, for the ones where form and details are exactly identical can you just copy the details object to make that obvious. If they diverge in the future the details can just be copied and pasted then modified.

client/src/components/resources/configs/pdx.js
```diff
@@ -0,0 +1,161 @@
...
+      'response_to_treatment',
+      'virology_status'
+    ]
+  },
+  {
+    clinical_tumor_information: [
+      'submitter_tumor_id',
+      'tissue_of_origin',
+      'tumor_type',
+      'speciment_tumor_tissue',
```
kurtwheeler, line 29:
```suggestion
      'specimen_tumor_tissue',
```

client/src/helpers/readableNames.js
```diff
@@ -1,36 +1,144 @@
...
+  purpose: 'Purpose',
+  relevant_mutations: 'Relevant Mutations',
+  response_to_qa_performed: 'Response to QA Performed',
+  response_to_standard_of_care: 'Response to Standard of Care',
+  response_to_treatment: 'Response to Treatment',
+  sequence_maps: 'Sequence Maps',
+  sex: 'Sex',
+  shipping_requirements: 'Shipping Requirements',
+  specific_markers: 'Specific Markers',
+  speciment_tumor_tissue: 'Specimen Tumor Tissue',
```
kurtwheeler, line 100:
```suggestion
  specimen_tumor_tissue: 'Specimen Tumor Tissue',
```

client/src/schemas/material.js
```diff
@@ -0,0 +1,156 @@
...
+    ethnicity: string(),
+    treatment_drug: string(),
+    treatment_protocol: string(),
+    prior_treatment_protocol: string(),
+    response_to_treatment: string(),
+    virology_statu: string(),
+    submitter_tumor_id: string(),
+    tissue_of_origin: string(),
+    tumor_type: string(),
+    speciment_tumor_tissue: string(),
```
kurtwheeler, line 92:
```suggestion
    specimen_tumor_tissue: string(),
```

client/src/components/resources/configs/pdx.js
```diff
@@ -0,0 +1,161 @@
...
```
kurtwheeler, line 104:
```suggestion
      'specimen_tumor_tissue',
```

client/src/components/resources/configs/protocol.js
```diff
@@ -0,0 +1,28 @@
+export const searchResult = ['description_or_abstract']
+
+export const details = [
+  {
+    basic_information: [
+      'protocol_name',
+      'description_or_abstract',
+      'additional_info'
```
kurtwheeler, line 8: Is abstract missing here?

client/src/helpers/readableNames.js
```diff
@@ -1,36 +1,144 @@
 export const readableNames = {
+  ADDGENE: 'Add Gene',
```
kurtwheeler, line 2: I think this should be `Addgene`. From their website:
```
Addgene is open and shipping!
```

client/src/schemas/material.js
```diff
@@ -0,0 +1,156 @@
...
+]
+
+export const importSources = [
+  'GEO',
+  'SRA',
+  'DBGAP',
+  'DATASET',
+  'PROTOCOLS_IO',
+  'ADDGENE',
+  'JACKSON_LAB',
```
kurtwheeler, line 20: In readable names this is `JACKSON_LABS`. Should they match?

client/src/schemas/material.js
```diff
@@ -0,0 +1,156 @@
...
+    description: string(),
+    number_samples: string(),
+    technology: string(),
+    platform: string()
+  }),
+  MODEL_ORGANISM: object({
+    description: string(),
+    genetic_background: string(),
+    zygostity: string(),
+    number_of_available_models: string(),
```
kurtwheeler, line 74: This shouldn't be an int?

client/src/schemas/material.js
```diff
@@ -0,0 +1,156 @@
...
+    growth_strain: string()
+  }),
+  PROTOCOL: object({
+    description: string(),
+    abstract: string(),
+    protocol_name: string()
+  }),
+  DATASET: object({
+    description: string(),
+    number_samples: string(),
```
kurtwheeler, line 66: This shouldn't be an int?

client/src/schemas/material.js
```diff
@@ -0,0 +1,156 @@
...
+  'ZIRC',
+  'OTHER'
+]
+
+const materialCategories = {
+  CELL_LINE: object({
+    cell_line_name: string(),
+    tissue: string(),
+    disease: string(),
+    number_of_availible_cell_lines: string(),
```
kurtwheeler, line 31: This shouldn't be an int?

client/src/schemas/material.js
```diff
@@ -0,0 +1,156 @@
...
+    tumor_grade: string(),
+    disease_stage: string(),
+    specific_markers: string(),
+    is_from_untreated_patient: string(),
+    tumor_sample_type: string(),
+    existing_model_explanatio: string(),
+    pdx_id: string(),
+    model_organism: string(),
+    model_strain_and_source: string(),
+    is_strain_immunized: string(),
```
kurtwheeler, line 103: This shouldn't be a bool?

client/src/schemas/material.js
kurtwheeler: I think every category has a field for the number of them, so I'm gonna stop commenting on each one.

client/src/schemas/material.js
```diff
@@ -0,0 +1,156 @@
...
+    virology_statu: string(),
+    submitter_tumor_id: string(),
+    tissue_of_origin: string(),
+    tumor_type: string(),
+    speciment_tumor_tissue: string(),
+    histology: string(),
+    tumor_grade: string(),
+    disease_stage: string(),
+    specific_markers: string(),
+    is_from_untreated_patient: string(),
```
kurtwheeler, line 97: This shouldn't be a bool?

client/src/schemas/material.js
```diff
@@ -0,0 +1,156 @@
...
+    is_strain_immunized: string(),
+    type_of_humanization: string(),
+    tumor_preparation: string(),
+    injection_type_and_site: string(),
+    treatment_for_engraftment: string(),
+    engraftment_rate: string(),
+    engraftment_tim: string(),
+    tumor_chracterization_technology: string(),
+    is_not_of_EBV_origin: string(),
+    is_passage_QA_performed: string(),
```
kurtwheeler, line 112: This shouldn't be a bool?

client/src/components/resources/configs/protocol.js
davidsmejia: it is supposed to show description 
but if that isnt defined it will show abstract 
and if that isnt defined it will say description is not specified or empty or something like that

client/src/schemas/material.js
davidsmejia: good catch my man

client/src/schemas/material.js
davidsmejia: When I talked it over with deepa, we thought units should be included which makes leaving it a string a good idea.

client/src/schemas/material.js
davidsmejia: this probably is safer to be an int, i think when we do review of the inputs we can modify it then (that will be pre-launch)

client/src/schemas/material.js
davidsmejia: same comment and above

client/src/schemas/material.js
davidsmejia: yes it should be

client/src/schemas/material.js
davidsmejia: yeah thats a good idea

client/src/components/resources/configs/cellLine.js
davidsmejia: yeah i think thats a good idea i can create an issue to simplify that stuff

Issue #54, opened 2020-03-26
EndToEndTest Set up organization with two users


kurtwheeler:
### Context

I've created a set of end to end tests for us to implement: https://github.com/AlexsLemonade/resources-portal/issues/49

### This test should test:

Two users create accounts, one creates an organization, adds the other user to it, grants approval permission, then lists a resource in it

#### The test should do that by making the following API calls:
1. Create account Prof
2. Create account Postdoc
3. Prof creates organization Lab with grant id.
4. Prof invites Postdoc to join Lab.
5. Postdoc accepts invitation to join Lab.
7. Prof lists resource under Lab.
8. Prof removes all notification settings.

#### This test should test that the following notifications occur and no more (see https://github.com/AlexsLemonade/resources-portal/issues/51):
- [ ] The Postdoc receives a notification that the Prof invited her to join her lab.
- [ ] The Prof is notified that Postdoc accepted her invitation.

dvenprasad:
Yeah...we want to notify what permissions they have once they accept invite/join request is approved. 

Also, when their permissions are modified.

PR #238, opened 2020-07-07
Organization with two users end to end test

Philinphiladelphia:
## Issue Number

#54 

This adds a test for the flow of creating a organization with two users.

## Types of changes

## Checklist


api/resources_portal/test/end_to_end/test_organization_with_two_users.py
```diff
@@ -0,0 +1,141 @@
...
+        # Create account Prof
+        self.client.get(get_mock_oauth_url([self.grant]))
+        prof = User.objects.get(pk=self.client.session["_auth_user_id"])
+
+        # Create account Postdoc
+        self.client.get(get_mock_oauth_url([]))
+        post_doc = User.objects.get(pk=self.client.session["_auth_user_id"])
+
+        # Prof creates organization Lab with grant id
+        # null members might be an issue?
```
kurtwheeler, line 54: I guess they weren't?

api/resources_portal/test/end_to_end/test_organization_with_two_users.py
```diff
@@ -0,0 +1,141 @@
...
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+        lab = Organization.objects.get(pk=response.data["id"])
+
+        # Prof invites Postdoc to join Lab
+        self.client.force_authenticate(user=prof)
+
+        invitation = OrganizationInvitation(
+            organization=lab, request_reciever=prof, requester=post_doc, invite_or_request="INVITE",
+        )
```
kurtwheeler, line 73: Can you construct this request with json, using the lab ID returned by the previous post (rather than fetching from the database)? These end to end tests are mocking the way the frontend will use the API, and it won't be able to construct objects this way or query the database.

PR #239, opened 2020-07-09
Modified attachment perms and updated tests

Philinphiladelphia:
## Issue Number

N/a

This PR does the following, pursuant of David's comments on the API:
- Uploads attachments before material-requests are made in the multiple_resources_requested_and_fulfilled end-to-end test.
- Adds field "owned_by_org" and "owned_by_user" to attachment.
- Changes perms on Attachments so they are based around ownership.
- Checks if the user is in an org which owns an attachment before PUTing it to a material-request.
- Modifies material-request and attachment unit tests to be in line with changes.
- Makes use of populate_dev_database consistent across all tests.
- Changed populate_dev_database so it creates ids dynamically as opposed to statically.
- Ensured that tests which use the search index reset it and subsequently restore it.

## Types of changes

## Checklist


api/resources_portal/management/commands/populate_dev_database.py
```diff
@@ -98,4 +98,4 @@ class Command(BaseCommand):
     help = "Populates the database with test data"
 
     def handle(self, *args, **options):
-        populate_test_database()
+        populate_dev_database()
```
kurtwheeler, line 101: :+1:

api/resources_portal/test/end_to_end/test_multiple_resources_requested_and_fulfilled.py
```diff
@@ -73,11 +73,41 @@ def test_create_account_and_list_resource(self, mock_auth_request, mock_record_r
...
+        irb1 = Attachment(
+            filename="executed_mta",
+            description="Executed transfer agreement for the material.",
+            s3_bucket="a bucket",
+            s3_key="a key",
+        )
+
+        irb2 = Attachment(
+            filename="executed_mta",
+            description="Executed transfer agreement for the material.",
```
kurtwheeler, line 86: The filename and descriptions say executed MTAs rather than IRB.

Philinphiladelphia: Good catch. Fixed.

api/resources_portal/models/attachment.py
```diff
@@ -34,6 +34,14 @@ class Meta:
     s3_bucket = models.CharField(max_length=255, blank=True, null=True)
     s3_key = models.CharField(max_length=255, blank=True, null=True)
 
+    owned_by_org = models.ForeignKey(
+        "Organization",
+        blank=False,
+        null=False,
```
kurtwheeler, line 40: Why is `null=False` for this field? Don't we want users to be able to upload attachments for their requests. What org are those owned by?

api/resources_portal/test/end_to_end/test_multiple_resources_requested_and_fulfilled.py
```diff
@@ -73,11 +73,41 @@ def test_create_account_and_list_resource(self, mock_auth_request, mock_record_r
         # Select two resources from PrimaryLab and request them, uploading two signed IRBs
         self.client.force_authenticate(user=requester)
 
+        # Upload IRBs
+        irb1 = Attachment(
+            filename="irb_attachment",
+            description="Institutional Board Review for the research in question.",
+            s3_bucket="a bucket",
+            s3_key="a key",
+        )
```
kurtwheeler, line 82: I'm surprised this works if `owned_by_org` isn't set.

api/resources_portal/models/attachment.py
Philinphiladelphia: Good point. Changed to True.

api/resources_portal/test/end_to_end/test_multiple_resources_requested_and_fulfilled.py
Philinphiladelphia: It did not! It has since been updated.

api/resources_portal/models/attachment.py
```diff
@@ -34,6 +35,18 @@ class Meta:
     s3_bucket = models.CharField(max_length=255, blank=True, null=True)
     s3_key = models.CharField(max_length=255, blank=True, null=True)
 
+    owned_by_user = models.ForeignKey(
+        "User", blank=False, null=True, on_delete=models.CASCADE, related_name="owned_attachments"
```
kurtwheeler, line 39: Do we ever want this to be null? I wouldn't think so.

api/resources_portal/models/attachment.py
```diff
@@ -53,3 +66,8 @@ def download_url(self):
             return "https://s3.amazonaws.com/" + self.s3_bucket + "/" + self.s3_key
         else:
             return None
+
+    def clean(self):
+        super(Attachment, self).clean()
+        if not (self.owned_by_org or self.owned_by_user):
+            raise ValidationError("Either owned_by_user or owned_by_org must be set on Attachment")
```
kurtwheeler, line 73: Hmm, I think it might make sense to always have `owned_by_user` be the user that uploaded the file, even if they did so from the organization side of things. Then this method won't do much any more.

api/resources_portal/test/factories.py
```diff
@@ -127,6 +127,7 @@ class Meta:
     description = "A file for testing"
     s3_bucket = "https://bucket-name.s3.region.amazonaws.com/keyname"
     s3_key = "s3 key"
+    owned_by_org = factory.SubFactory(OrganizationFactory)
```
kurtwheeler, line 130: I think we should also have an `owned_by_user` field here.

api/resources_portal/test/utils.py
```diff
@@ -58,3 +61,10 @@ def get_mock_oauth_url(grants):
         url += f"&grant_id={grant.id}"
 
     return url
+
+
+def restore_database():
+    for name, obj in inspect.getmembers(models):
+        if inspect.isclass(obj):
+            for model in obj.deleted_objects.all():
+                model.save()
```
kurtwheeler, line 70: What's this save doing? It doesn't seem to be modifying anything.

api/resources_portal/test/views/test_attachment.py
```diff
@@ -165,15 +148,22 @@ def test_delete_request_from_admin_succeeds(self):
...
 
-    def test_delete_request_from_user_succeeds(self):
+    def test_delete_request_from_user_in_org_succeeds(self):
         self.client.force_authenticate(user=self.user)
 
         response = self.client.delete(self.url)
 
         self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
 
+    def test_delete_request_from_user__not_in_org_fails(self):
```
kurtwheeler, line 158:
```suggestion
    def test_delete_request_from_user_not_in_org_fails(self):
```

api/resources_portal/test/views/test_material_request.py
```diff
@@ -133,17 +139,33 @@ def test_put_request_from_sharer_updates_a_material_request(self):
...
+        requester_org = OrganizationFactory()
+        requester_org.members.add(self.request.requester)
+        requester_org.save()
+
+        irb_attachment = AttachmentFactory(owned_by_org=requester_org)
 
         self.material_request_data["irb_attachment"] = irb_attachment.id
-        self.material_request_data["requester_signed_mta_attachment"] = AttachmentFactory().id
+        self.material_request_data["requester_signed_mta_attachment"] = AttachmentFactory(
+            owned_by_org=requester_org
```
kurtwheeler, line 150: I think we want to be testing `owned_by_user` in this method rather than `owned_by_org`

api/resources_portal/views/attachment.py
```diff
@@ -21,45 +22,37 @@ class Meta:
...
 
 
-class CanApproveRequestsOrIsRequesterOrIsAdminUser(BasePermission):
+class OwnsAttachmentOrIsRequesterOrIsAdmin(BasePermission):
     def has_object_permission(self, request, view, obj):
         return (
-            MaterialRequest.objects.filter(requester=request.user, is_active=True).exists()
-            or user_has_perm_on_active_material_request(request.user, "approve_requests")
+            request.user in obj.owned_by_org.members.all()
+            or user_is_requester_on_active_material_request(request.user, obj.owned_by_org)
```
kurtwheeler, line 55: I think we can use `request.user == obj.owned_by_user or request.user in obj.owned_by_org.members.all() or request.user.is_staff` and call it `OwnsAttachmentOrIsAdmin`, and the we don't need to worry about whether the user is the requester or not, and can nix the `user_is_requester_on_active_material_request` function.

api/resources_portal/views/material_request.py
```diff
@@ -87,15 +88,50 @@ def send_material_request_notif(notif_type, request):
...
+        send_material_request_notif("TRANSFER_FULFILLED", request, request.requester)
     else:
         return
 
 
+def user_owns_attachment(attachment, user):
+    if attachment.owned_by_org:
+        return user in attachment.owned_by_org.members.all()
+    else:
+        return False
```
kurtwheeler, line 104: Why doesn't this function check `attachment.owned_by_user`?

api/resources_portal/views/material_request.py
```diff
@@ -87,15 +88,50 @@ def send_material_request_notif(notif_type, request):
...
+
+
+def user_in_attachment_org(attachment, user):
+    if attachment.owned_by_user:
+        return user == attachment.owned_by_user
```
kurtwheeler, line 109: This check doesn't seem to match the function name. I think we can probably remove this function and just add this check to `user_owns_attachment` though.

api/resources_portal/views/attachment.py
```diff
@@ -96,13 +95,6 @@ def get_permissions(self):
     # because we want the path to have its id in it.
     @transaction.atomic
     def create(self, request, *args, **kwargs):
-        if not (
-            MaterialRequest.objects.filter(requester=request.user, is_active=True).exists()
-            or user_has_perm_on_active_material_request(request.user, "approve_requests")
```
kurtwheeler, line 101: These three places are the only places this function gets used right? It seems like you should be able to delete it then.

api/resources_portal/models/attachment.py
kurtwheeler: This error now seems unnecessary.

api/resources_portal/views/material_request.py
```diff
@@ -93,15 +94,43 @@ def send_material_request_notif(notif_type, request):
...
+        return user in attachment.owned_by_org.members.all()
+    else:
+        return False
+
+
+# Adds an attachment to a material request, checking that the current user is in the org that uploaded the attachment.
+def add_attachment_to_material_request(material_request, attachment, attachment_type, user):
+    if not (user_in_attachment_org(attachment, user) or attachment.owned_by_user == user):
+        raise PermissionDenied(
+            detail=f"The current user is not authorized for the specifified attachment of type {attachment_type}."
```
kurtwheeler, line 117:
```suggestion
            detail=f"The current user is not authorized for the specified attachment of type {attachment_type}."
```

api/resources_portal/views/attachment.py
Philinphiladelphia: Good call. Deleted.

Issue #217, opened 2020-06-22
Set up S3 for attachments


kurtwheeler:
### Context

#211 adds a route for attachments.

### Problem or idea

However it doesn't actually put them into S3 because I need to configure that.

### Solution or next step

Hook the attachments route up to S3 and test that files can be uploaded with it.

davidsmejia:
For local envs could we have the endpoints instead of sending to s3, save the file locally. Would probably make sense to add a folder to the repo and git ignore it.

Issue #209, opened 2020-06-08
Upload files with attachments


kurtwheeler:
### Context

We have an attachments route, but we need a way to actually pass along the file along.

### Problem or idea

Not quite sure how Django passes files, but it's gotta be supported like photos.

We should have the user upload to our server, then we upload to S3.

### Solution or next step

Allow files to be uploaded.

PR #240, opened 2020-07-13
File uploads

kurtwheeler:
## Issue Number

#209 
#217 

This knocks out both local file uploads/downloads along with doing it in the cloud using s3 as well. 

## Types of changes

## Functional tests

I've tested this locally as well as in my dev stack by disabling auth and using:
```
curl -v -F "description=heheh" -F file=@api/dev_data/nerd_sniping.png http://52.5.233.92/v1/attachments
```

## Checklist


api/resources_portal/middleware/oauth.py
```diff
@@ -51,6 +54,8 @@ def __call__(self, request):
             # get user orcid info
             response = requests.post(OAUTH_URL, data=data, headers={"accept": "application/json"})
             response_json = response.json()
+            if "orcid" not in response_json:
```
Philinphiladelphia, line 57: Nice change!

api/resources_portal/test/views/test_attachment.py
```diff
@@ -36,19 +40,40 @@ def setUp(self):
         self.admin = UserFactory()
         self.admin.is_staff = True
 
+        # Cleanup the attachments test directory so there's no files
+        # from previous tests:
+        for directory_name in os.listdir(settings.LOCAL_FILE_DIRECTORY):
```
Philinphiladelphia, line 45: Do we want to convert existing end-to-end tests to use real files? I imagine we do. If so, it would be great if this could be made into a function in the "mocks.py" file (renamed to utils.py in my update-attachments PR). This would allow it to be reused.

infrastructure/api-configuration/dev-secrets
```diff
@@ -1,2 +1,5 @@
 DJANGO_SECRET_KEY=THIS_IS_NOT_A_SECRET_DO_NOT_USE_IN_PROD
 DATABASE_PASSWORD=THIS_IS_NOT_A_SECRET_DO_NOT_USE_IN_PROD
+
+OAUTH_CLIENT_SECRET=1aaa0a94-5605-4cd8-95a4-6724c2f7fd42
+OAUTH_URL=https://sandbox.orcid.org/oauth/token
```
Philinphiladelphia, line 5: Will this need to be changed manually off of sandbox when we go to full deployment?

Philinphiladelphia: Ah- never mind. Just read below.

Issue #241, opened 2020-07-14
SRA Importer


kurtwheeler:
### Context

Some resources can be imported. SRA datasets is one of these.

### Problem or idea

I imagine the way this will work is that the user will provide a URL/accession (they're kinda interchangable, @dvenprasad which do you prefer we ask for?). This URL/accession will be sent to the backend at a route like `v1/materials/import`. The backend will then use that URL/accession to hit an API and import a material from it.

### Solution or next step

Look at the material model and populate any fields in it that can be populated using the SRA API. See https://github.com/AlexsLemonade/refinebio/blob/dev/foreman/data_refinery_foreman/surveyor/sra.py for code that extracts a lot of fields from their XML. There may be an easier API to use to get the more limited set of fields needed for this project.

dvenprasad:
The form leaves it up to the user which one they want to input. Is that okay or do we need to restrict them to either URL or Accession?

kurtwheeler:
Nope, we can accept either! It should be easy to pull the accession out of the URL.

PR #248, opened 2020-07-20
Add sra import

Philinphiladelphia:
## Issue Number

#241 

This adds a route for importing materials, implementing the SRA import option. Others will be added in subsequent stories. This PR also adds tests for the import-sra route and docs for the import endpoint.

## Types of changes

## Functional tests

Functional tests can be found in test>views>test_import_sra.py.

## Checklist


api/resources_portal/urls.py
```diff
@@ -89,3 +90,7 @@
     re_path(r"^$", RedirectView.as_view(url=reverse_lazy("api-root"), permanent=False)),
     path("v1/search/", include(search_router.urls)),
 ] + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
+
+urlpatterns.append(
+    path("v1/materials/import/", ImportViewSet.as_view({"post": "create"}), name="import")
```
kurtwheeler, line 95: Any reason to not use `router.register` for this one?

api/resources_portal/views/import_materials.py
```diff
@@ -0,0 +1,88 @@
...
+    This function returns a Response object containing the json representation of the newly-created material object
+    made using SRA data.
+    """
+    metadata = gather_all_metadata(run_accession)
+
+    # note for the PR: currently, I use "library-construction-protocol" to populate the description. This is a technical description of the steps taken to process the material.
+    # Should I use the study abstract to populate it instead?
+
+    # One more question for PR. I was not able to find a way to retrieve "num_samples", "technology", "pre_print_doi", or "pre_print_title"
+    # from the SRA api. Are these neccessary to retrieve automatically? If so, I may need to pair with someone to find out out how to get those.
```
kurtwheeler, line 21: For num_samples, take a look at https://github.com/AlexsLemonade/refinebio/blob/dev/foreman/data_refinery_foreman/surveyor/sra.py#L578. The number of runs is the number of samples for our purposes. If you do something similar to that and then do `len(accessions_to_run` that would be good for `num_samples`.

For technology, I see `'library_strategy': 'RNA-Seq'` in the metadata. @cgreene do you think this field will be what we want for technology?

I'm not actually sure how to get pre-print info. @cgreene is probably the person to ask about this as well.

api/resources_portal/views/import_materials.py
```diff
@@ -0,0 +1,88 @@
...
+
+def import_sra(run_accession, organization, grant, user):
+    """
+    This function returns a Response object containing the json representation of the newly-created material object
+    made using SRA data.
+    """
+    metadata = gather_all_metadata(run_accession)
+
+    # note for the PR: currently, I use "library-construction-protocol" to populate the description. This is a technical description of the steps taken to process the material.
+    # Should I use the study abstract to populate it instead?
```
kurtwheeler, line 18: Probably another good question for @cgreene, but I would think the study abstract because these datasets will be at the study/experiment level rather than sample level.

api/resources_portal/test/views/test_import_sra.py
```diff
@@ -0,0 +1,109 @@
...
+from resources_portal.test.factories import GrantFactory, OrganizationFactory, UserFactory
+
+
+class ImportSRATestCase(APITestCase):
+    """
+    Tests importing via SRA.
+    """
+
+    def setUp(self):
+        self.test_accession = "SRR4199304"
```
kurtwheeler, line 15: I think we want to use accessions that start with `SRP` rather than `SRR` because they correspond to studies rather than runs.

api/resources_portal/views/import_materials.py
cgreene: I think study/experiment abstract - yea.

api/resources_portal/urls.py
Philinphiladelphia: Only that I don't know how I would include multiple parameters in the URL if I did it with router.register. Do you know how to do that?

api/resources_portal/views/import_materials.py
Philinphiladelphia: Cool. I added these two parameters, and left out pre-print info for now.

api/resources_portal/views/import_materials.py
Philinphiladelphia: Changed it to abstract.

api/resources_portal/test/views/test_import_sra.py
Philinphiladelphia: Good call. I reworked the api calls to do this.

api/resources_portal/views/import_materials.py
```diff
@@ -0,0 +1,100 @@
...
+        if request.user not in organization.members.all():
+            return JsonResponse(
+                {"error": f"The user is not a member of organization id {organization.id}"},
+                status=403,
+            )
+
+        import_type = request.data["import_type"]
+
+        if import_type == "SRA":
+            return import_sra(request.data["run_accession"], organization, grant, request.user)
```
kurtwheeler, line 96: We're expecting study accessions now, not run_accessions. I think the field name should be changed to reflect that, but it also means that `gather_all_metadata` needs to be updated to start with study accession instead of run accession.

api/resources_portal/views/import_materials.py
```diff
@@ -0,0 +1,100 @@
...
+    if metadata == {}:
+        return JsonResponse(
+            {"error": "No data was found for the provided accession code"}, status=404
+        )
+
+    # note for the PR: currently, I use "library-construction-protocol" to populate the description. This is a technical description of the steps taken to process the material.
+    # Should I use the study abstract to populate it instead?
+
+    # One more question for PR. I was not able to find a way to retrieve "num_samples", "technology", "pre_print_doi", or "pre_print_title"
+    # from the SRA api. Are these neccessary to retrieve automatically? If so, I may need to pair with someone to find out out how to get those.
```
kurtwheeler, line 26: These comments can be removed now right?

api/resources_portal/views/import_sra_utils.py
```diff
@@ -18,6 +18,28 @@ class UnsupportedDataTypeError(Exception):
...
+    "SRS000001, SRS000002, SRS000003-SRS000008"
+    You can count the non-consecutive elements, but the ranges must be parsed.
+    """
+
+    num_samples = 0
+
+    srs_list = srs_string.split(",")
+
+    for srs in srs_list:
+        srs = srs.replace("SRS", "")
```
kurtwheeler, line 33: Wait, we want number of runs, not samples (I mentioned that in https://github.com/AlexsLemonade/resources-portal/pull/248#discussion_r457632161). I can understand why this is confusing, because like, why wouldn't we want to count the sample accessions for number of samples? However, an `SRS` accession is generated for each chunk of organism that is collected, which may then be run through the machine multiple times. Each time it gets run through the machine, an `SRR` accession is generated. In refinebio we've used `samples` to refer to runs, since each one is a new column of gene expression data.

I think for this project the same holds true, since if you download a dataset with 4 samples and then see 16 columns it'd probably be more surprising than finding out that a dataset claiming to have 16 samples was actually creating using only 4 chunks of organism. Can you confirm this @dvenprasad ?

api/resources_portal/views/import_materials.py
kurtwheeler: Ah, I just noticed that you already reworked `gather_all_metadata` to use study accessions but just didn't change the field name for the API.

api/resources_portal/views/import_sra_utils.py
dvenprasad: Yep. that sounds reasonable to me.

api/resources_portal/urls.py
kurtwheeler: You said in chat:

> I'm not sure how to register "material/import" without the system wanting it to be "material/<id>/import"

I agree that's a good reason. Can you just change the name to `materials-import` though?

api/resources_portal/views/import_sra_utils.py
```diff
@@ -30,7 +30,7 @@ def _get_number_of_samples(srs_string):
     srs_list = srs_string.split(",")
 
     for srs in srs_list:
-        srs = srs.replace("SRS", "")
+        srs = srs.replace("SRR", "")
```
kurtwheeler, line 33: Can you change all variables with `srs` in them to `srr`?

Issue #243, opened 2020-07-14
protocols.io importer


kurtwheeler:
### Context

Some resources can be imported. protocols.io is one of these.

### Problem or idea

I imagine the way this will work is that the user will provide a URL/accession. This URL/accession will be sent to the backend at a route like `v1/materials/import`. The backend will then use that URL/accession to hit an API and import a material from it.

### Solution or next step

Look at the material model and populate any fields in it that can be populated using the protocols.io API. We've never pulled data/metadata from them before so you'll have to just check out https://www.protocols.io/developers and https://apidoc.protocols.io/

PR #268, opened 2020-07-27
Add protocol io importer

Philinphiladelphia:
## Issue Number

#243 

This adds the importer for protocols.io materials. It also updates the docs for GEO and protocols.io imports and adds tests for protocol.io imports.

Note: I was not able to find a way to get pubmed id/publication title anywhere on a protocols.io website. If you think I should do some more digging, just let me know and I'll see if there's anything else I can find.

## Types of changes

## Checklist


docs/api/import_materials.md
```diff
@@ -63,3 +64,138 @@ Content-Type application/json
...
+**Request**:
+
+`POST` `/materials/import/`
+
+Parameters:
+
+Name                           | Type   | Description
+-------------------------------|--------|---
+import_type                    | string | The type of resource being imported. Options include: ["SRA", "GEO", "PROTOCOLS.IO"].
+study_accession                | int    | The accession code for the study represeting the SRP material, in the form "SRP123456".
```
kurtwheeler, line 79: This is for SRA, however I'm not actually sure we need a separate section for GEO. I think having one section titled `Import a Dataset` would be sufficient because SRA and GEO aren't really all that different from the user's perspective.

docs/api/import_materials.md
```diff
@@ -11,14 +11,15 @@ Parameters:
 
 Name                           | Type   | Description
 -------------------------------|--------|---
-import_type                    | string | The type of resource being imported. Options include: ["SRA"].
+import_type                    | string | The type of resource being imported. Options include: ["SRA", "GEO", "PROTOCOLS.IO"].
 study_accession                | int    | The accession code for the study represeting the SRP material, in the form "SRP123456".
```
kurtwheeler, line 15:
```suggestion
study_accession                | int    | The accession code for the study representing the material, in the form "SRP123456" for SRA or "GSE123456" for GEO.
```

docs/api/import_materials.md
```diff
@@ -63,3 +64,138 @@ Content-Type application/json
...
+**Request**:
+
+`POST` `/materials/import/`
+
+Parameters:
+
+Name                           | Type   | Description
+-------------------------------|--------|---
+import_type                    | string | The type of resource being imported. Options include: ["SRA", "GEO", "PROTOCOLS.IO"].
+study_accession                | int    | The accession code for the study represeting the SRP material, in the form "SRP123456".
```
kurtwheeler, line 153: This should be updated to reflect what you provide protocols.io. I think we should maybe change `study_accession` to be something more generic like `accession` or `material_accession` so that it can work for every type of material we ever write an importer for. (Post MVP there may be more than these three.)

docs/api/import_materials.md
Philinphiladelphia: Gotcha. Just merged the two sections.

docs/api/import_materials.md
Philinphiladelphia: Added.

docs/api/import_materials.md
Philinphiladelphia: For protocols.io, you provide a DOI- I just changed it to that.

davidsmejia:
Does the import type here need to match what is defined in the materials model?

`PROTOCOLS_IO`

Issue #165, opened 2020-05-07
Configure logger


Philinphiladelphia:
### Context

We currently have no infrastructure to print out custom logging messages from resources-portal.

### Problem or idea

We should add and configure a logger.

### Solution or next step

We can implement this using the "logging" module in python. A good place to start would be the Notifications model email message.

PR #269, opened 2020-07-29
Add logger

Philinphiladelphia:
## Issue Number

#165 

This adds and configures a logger to output all messages.

## Types of changes

## Functional tests

There are no tests for this currently.

## Checklist


api/resources_portal/config/logging.py
```diff
@@ -0,0 +1,77 @@
...
+
+            @retry(stop_max_attempt_number=3)
+            def retrieve_instance_id():
+                return requests.get(os.path.join(METADATA_URL, "instance-id")).text
+
+            INSTANCE_ID = retrieve_instance_id()
+        else:
+            INSTANCE_ID = "local"
+
+    return INSTANCE_ID
```
kurtwheeler, line 31: I actually don't think we need this one in this project. We only have the one API instance so it won't help in narrowing that down like it does in refinebio.

api/resources_portal/config/common.py
```diff
@@ -50,6 +50,7 @@ class Common(Configuration):
     ROOT_URLCONF = "resources_portal.urls"
     SECRET_KEY = os.getenv("DJANGO_SECRET_KEY")
     WSGI_APPLICATION = "resources_portal.wsgi.application"
+    RUNNING_IN_CLOUD = False
```
kurtwheeler, line 53: We don't need this any more.

Philinphiladelphia: Ahh, true.

Issue #283, opened 2020-08-03
View all of a users requests 


davidsmejia:
While this is possible we need to check if permissions filter out the results - and create a test for that.

kurtwheeler:
So right now responses from the users endpoint will contain a `requests` field that will only have requests that they made along with an `assignments` field that is the requests where they're the `assigned_to` user. I think this covers what we want right?

PR #291, opened 2020-08-03
Filters list material request endpoint to only requests the user can view

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/resources-portal/issues/283

Turns out https://www.django-rest-framework.org/api-guide/permissions/#limitations-of-object-level-permissions is a thing. Basically it doesn't work for list endpoints. I checked and this is actually the only place we were using it for lists.

## Types of changes

## Functional tests

The unit test I added was failing before I got this working.

## Checklist

### DIFF
--- a/api/resources_portal/test/views/test_material_request.py
+++ b/api/resources_portal/test/views/test_material_request.py
@@ -103,10 +103,21 @@ class TestMaterialRequestListTestCase(APITestCase):
         response = self.client.post(self.url, self.material_request_data, format="json")
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
 
-    def test_get_request_succeeds(self):
+    def test_get_request_from_sharer_succeeds(self):
         self.client.force_authenticate(user=self.sharer)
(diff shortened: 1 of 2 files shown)


api/resources_portal/test/views/test_material_request.py
```diff
@@ -103,11 +103,22 @@ def test_post_request_from_unauthenticated_forbidden(self):
...
         self.client.force_authenticate(user=self.sharer)
         response = self.client.get(self.url)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
 
+    def test_get_request_from_requester_succeeds(self):
+        self.client.force_authenticate(user=self.request.requester)
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_get_request_filters(self):
```
Philinphiladelphia, line 116: Another test where the filter returns one or more requests would be good. You could even just add `self.assertEqual(response.json()["count"], 1)` at the end of the previous two tests if you wanted.

Issue #288, opened 2020-08-03
Material Request needs Verified state


davidsmejia:
Add verified status option

Issue #284, opened 2020-08-03
Material Request: assigned_to


davidsmejia:
`material_request.assigned_to` needs to be made not read only aka it should be editable

PR #293, opened 2020-08-05
Makes assigned_to read-only for requester, add verified state

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/resources-portal/issues/284
https://github.com/AlexsLemonade/resources-portal/issues/288

This handles two issues and adds tests for both of them. The first is making assigned_to editable by org members with permissions to approve requests. The other is to add the VERIFIED_FULFILLED state to requests, and only allow a transition to it by the requester if the material is in the `FULFILLED` state.

## Types of changes

## Functional tests

I've added two unit tests for each issue.

## Checklist


kurtwheeler:
Interesting, it did want to make a migration, but somehow that wasn't actually required for the tests to pass?

Issue #287, opened 2020-08-03
Make sure users cannot remove their only grant.


davidsmejia:
Make sure the front end cant remove their only grant - we can do this on the FE but we should also prevent it from happening via the API in case of a bug

PR #295, opened 2020-08-05
Prevents last grant from being deleted/disassociated from a user.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/resources-portal/issues/287

This makes it so the last grant a user has cannot be deleted/disassociated.

## Types of changes

## Functional tests

I've added unit tests to cover this.

## Checklist


davidsmejia:
I think we should close this and instead remove support for post and delete on the grants endpoint until the future 😞 sorry

PR #302, opened 2020-08-11
Personal org and one user per grant

kurtwheeler:
## Issue Number

I don't think we made an issue for it

We need to be able to easily tell which org is the personal org for each user.

We also want each grant to only have a single user for now.

We want the personal org to be pretty transparent, so we want its grants and its owner's grants to stay in sync. That's what this was going to be about, but changing grants to only having one user ended up being a bigger change than I expected so that's mostly what this PR is about. I'll keep things in sync in a subsequent PR.

## Types of changes

## Functional tests

I got the unit tests working again!!!

## Checklist


api/resources_portal/test/factories.py
```diff
@@ -26,6 +26,9 @@ class Meta:
     is_active = True
     is_superuser = False
     is_staff = False
+    # This can get recurisve, so if someone needs it they can
```
Philinphiladelphia, line 29:
```suggestion
    # This can get recursive, so if someone needs it they can
```

PR #303, opened 2020-08-12
Transfer materials to personal organization when deleting grant

kurtwheeler:
## Issue Number

N/A came up in a discussion on Monday.

We decided that when a user disassociates their grant from an organization, that they should have an option to transfer the materials on that grant to their personal organization. This implements that with a `transfer=true` query param on the `organizations/<id>/grants` routes.

## Types of changes

## Functional tests

I made a unit test to test this.

## Checklist


api/resources_portal/test/views/test_organization_grant.py
```diff
@@ -137,6 +137,37 @@ def test_delete_request_disassociates_a_material(self):
...
+        grant.refresh_from_db()
+        self.assertNotIn(grant, self.organization1.grants.all())
+        # Verify that the grant was not deleted, just its relationship
+        grant = Grant.objects.get(pk=grant.id)
+
+        for material in grant.materials.all():
+            self.assertEqual(user.personal_organization, material.organization)
+
     def test_delete_fails_if_not_grant_owner(self):
         self.client.force_authenticate(user=self.organization1.owner)
```
davidsmejia, line 172:
```suggestion
        self.client.force_authenticate(user=self.organization2.owner)
```

davidsmejia: the user here is from org1 and so is the grant, we could do this or add the grant to org2

PR #306, opened 2020-08-13
Change operation of username field

Philinphiladelphia:
## Issue Number

N/A

This changes the username field on User to  be computed, as we don't ask the user to provide it. It also adds the field "full_name", as requested by David to make life on the frontend a bit easier. 

Question: should these tests get merged? They're for a pretty static functionality, so running them every time might not be necessary. They could be removed.

Supporting changes:
- Modify dev_data to use usernames in line with the computed ones
- Added full_name to user view
- Remove "POST" section for user docs. More info on user registration will be coming in a subsequent PR.

## Types of changes

## Functional tests

Tests at /test/models/test_user.py

## Checklist

## Screenshots


api/dev_data/users.json
```diff
@@ -32,8 +32,8 @@
     },
     {
       "username": "SecondaryProf",
-      "first_name": "Secundus",
-      "last_name": "Profarius",
+      "first_name": "Secondary",
+      "last_name": "Prof",
       "published_name": "Dr. Secundus Profarius",
```
kurtwheeler, line 37: These published names no longer match, not that it matters THAT much.

PR #320, opened 2020-08-19
adds useIsClient

davidsmejia:
## Issue Number

N/A This is a hook for deferred rendering.


client/src/hooks/useIsClient.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+
+// Some things are only available on the cliet like window and localStorage
```
davidsmejia, line 3:
```suggestion
// Some things are only available on the client like window and localStorage
```

Issue #315, opened 2020-08-17
Index archived materials with lowest priority for search


kurtwheeler:
### Context

Search should return archived materials, but at the bottom.

### Problem or idea

Elasticsearch has priorities for fields. There should be some way to set this up to put all archived materials at the bottom.

### Solution or next step

Make archived materials appear at the end of the search list.

PR #321, opened 2020-08-19
Add computed fields and order based on is_archived

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/resources-portal/issues/315

I'm not sure why everything is broken. I think it all went wrong when I tried to add is_archived to elasticsearch.

Other than that I added is_active, action_required_sharer, and action_required_requester to the model/view for material request.

I also change PENDING -> OPEN.

## Types of changes

## Functional tests

The unit tests cover this well enough I think!

## Checklist


api/resources_portal/models/material_request.py
```diff
@@ -64,14 +64,38 @@ class Meta:
...
+    def is_active(self):
+        return self.status in ["OPEN", "APPROVED", "IN_FULFILLMENT"]
+
+    @property
+    def requires_action_sharer(self):
+        if not self.is_active:
+            return False
+
+        if self.status == "APPROVED":
+            return not self.requires_action_requester()
```
davidsmejia, line 83: does this need to be executed as a function?

api/resources_portal/views/document_views.py
```diff
@@ -253,6 +262,7 @@ class MaterialDocumentView(DocumentViewSet):
 
     # Define ordering fields
     ordering_fields = {
+        "is_archived": "is_archived",
```
davidsmejia, line 265: did you try `-is_archived`?

kurtwheeler: Hmm, that might be necessary, but the current issue I'm running into is that any materials with `is_archived` are being filtered out of the result set, not misordered.

davidsmejia:
Just has one small conflict with `fulfillment_notes`

PR #322, opened 2020-08-19
Add model, view, and tests for fulfillment notes.

kurtwheeler:
## Issue Number

Note sure there is one

Sharers should be able to add multiple notes about fulfillment, tracked by date. This requires a model, which needs a view. This adds it all and tests it.

## Types of changes

## Functional tests

Unit tests!!

## Checklist


api/resources_portal/views/fulfillment_note.py
```diff
@@ -0,0 +1,79 @@
...
+        if self.action == "list":
+            return FulfillmentNoteSerializer
+
+        return FulfillmentNoteDetailSerializer
+
+    def get_permissions(self):
+        if self.action == "list":
+            permission_classes = [IsAdminUser]
+        else:
+            permission_classes = [IsAuthenticated, IsSharer]
```
davidsmejia, line 57: Can the requester see this as well?

davidsmejia: Or I'm guessing they can see it when attached to a material request

Issue #325, opened 2020-08-24
Exception when PATCH sent to /attachments/[id]


davidsmejia:
### Context

Was having an issue updating an attachment looks like in the permissions check.

I think this should be something like `obj.owned_by_org in request.user.organizations.all()`
Which would mean it would pass if the user is in an organization that the attachment belongs to.

```AttributeError: 'NoneType' object has no attribute 'members'```

https://github.com/AlexsLemonade/resources-portal/blob/dev/api/resources_portal/views/attachment.py#L58

If this tracks @kurtwheeler could you take a look into this as well:

I think this also exposes a need to pass an organization id when creating an attachment. Since this is my Personal org which I am owner but not a member of.

If I wanted to update this as a member of an org I don't see how the attachment would know it is kosher unless that org_id was set so we should probably require it (or check every relationship ie attachment->[sequence_map_for]->organization->members + owner.

kurtwheeler:
> attachment->[sequence_map_for]->organization->members + owner

I think is maybe the right thing to do. I wonder if we should just insist that owners are always members of the org. I think that might be simpler in a lot of ways and it makes more sense to me semantically.

PR #329, opened 2020-08-24
[WIP] Fix organization permissions for attachments

kurtwheeler:
## Issue Number

#325 

I'm not entirely sure if this does it all or not, but it's a good start. The owner of an organization should always be a member so this just ensures that is the case after every save.

## Types of changes

## Functional tests

None yet.

## Checklist


api/resources_portal/models/material.py
```diff
@@ -91,3 +93,30 @@ def has_publication(self):
...
+def fix_attachment_organizations(
+    sender, instance=None, created=False, update_fields=None, **kwargs
+):
+    """If the organziation changes, update the attachment.
+    """
+    if created or not instance:
+        # Nothing to do during creation.
+        return
+
+    attachments_needing_update = instance.sequence_maps.exclude(owned_by_org__isnull=True).exclude(
```
kurtwheeler, line 108: I can't remember why I excluded orgs attachments that had null orgs. If they're associated with a material then they should be owned by the same org as the material right? I think I should change this.
```suggestion
    attachments_needing_update = instance.sequence_maps.exclude(
```

PR #337, opened 2020-08-26
Get user email from orcid

Philinphiladelphia:
## Issue Number

N/a

This allows us to get user emails from ORCID. To accomplish this, the following changes were made:
- The 'Enter Email' step was moved to after the create account step and modified
- The /auth endpoint was split into the /login, /create-account and /orcid-credentials endpoints
- Tests were written for these three endpoints

Supporting changes:
- Removed /auth tests
- Fixed a small bug in the "user kidnapping" code

## Types of changes

## Functional tests

I will write tests for the new endpoints.

## Checklist


client/src/hooks/useCreateUser.js
```diff
@@ -18,22 +18,59 @@ export const useCreateUser = (
...
+      needsSave = true
+    }
+
+    if (grants && !createUser.grants) {
+      createUser.grants = grants
+      needsSave = true
+    }
+
+    if (needsSave) {
+      save()
```
davidsmejia, line 45: does this need to be in a useEffect and then set to false after?

Philinphiladelphia: Looks like it does not need to be in there. I added the lines setting it to "false" as well.

davidsmejia:
Hey this is looking good, I was a little confused about setting the redirectUrl in getInitialProps, seems as though that should only be set when someone clicks the login/create button.

There are three senarios we want to support:
A. The user is sent the login url with various information about creating a grant recipients account.
 - Here we want to redirect them to the create about flow so they can verify their grant info
 - They should always end up at the same end page when they are logged in. create team or list resource step)
B. A general user who has searched a material 
 - They are not a member so they want to create an account
 - We capture their location on the site and redirect them when they come back
C. A user signing into a new machine or their token has expired
 - pretty much the same as above we should redirect them where they were

You could either store the values in localStorage or send them along to orcid if it isn't an issue. If you want to pass it to orcid I would still get generate the url from the client.

This might have been too verbose but I see that the create account flow works, just wanted to make sure that the redirect url wasn't dependent on a server page load because getInitialProps only runs when doing SSR. So if you were to go to /this-page then click a button to go to the homepage, getInitialProps isn't executed before the homepage loads.

PR #366, opened 2020-09-01
updated modals

Philinphiladelphia:
## Issue Number

N/a

This combines all account creation modals into one configurable object. It also implements an option for non-dismissable modals.

## Types of changes

## Functional tests

none

## Checklist


client/src/components/CreateOrLogin.js
```diff
@@ -0,0 +1,140 @@
...
+            href="https://orcid.org/"
+            label="Learn more at orcid.org"
+          />
+        </Box>
+      </Box>
+    </Box>
+  )
+}
+
+export const CreateOrLoginModal = ({ title, includeSignInSection }) => {
```
davidsmejia, line 121: Could you call this `CreateOrLoginButton`

Philinphiladelphia: Yep!

davidsmejia:
Very nice. Just the one note about the Name

client/src/components/CreateOrLogin.js
davidsmejia: Sorry, could you name it `CreateAccountLoginButton` and then name the option `showSignIn` and set a default to the most common

davidsmejia: `title` should also have a default that corresponds to the `showSignIn` default

davidsmejia:
Could you also export `CreateOrLogin` and then update the filename to match the default export name too

client/src/components/Modal.js
```diff
@@ -2,28 +2,32 @@ import Icon from 'components/Icon'
...
+    }
+  }
   return (
     <Box>
       {showing && (
-        <Layer
-          onEsc={() => setShowing(false)}
-          onClickOutside={() => setShowing(false)}
-        >
+        <Layer onEsc={dismissModal} onClickOutside={dismissModal}>
```
davidsmejia, line 14: should these be dependent on `nondismissable`

davidsmejia: nevermind that looks good

Philinphiladelphia: They are, as part of the dismissModal function.

PR #381, opened 2020-09-03
Cleanup for create-account and redirect if there are no grants

Philinphiladelphia:
## Issue Number

N/a

This cleans up localstorage when you navigate to another page and redirects users hwo have no grants going to the /create-account page.

## Types of changes

## Functional tests

None

## Checklist

### DIFF
--- a/client/src/hooks/useCreateUser.js
+++ b/client/src/hooks/useCreateUser.js
@@ -30,6 +30,12 @@ export const useCreateUser = (
   } = React.useContext(CreateUserContext)
   const router = useRouter()
 
+  const handleRouteChangeStart = (url) => {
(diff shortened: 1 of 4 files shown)


client/src/hooks/useCreateUser.js
```diff
@@ -48,6 +51,16 @@ export const useCreateUser = (
     needsSave = false
   }
 
+  React.useEffect(() => {
```
davidsmejia, line 54: Could you define something like this here and remove the guard from the context?
```javascript
const handleChangeStart = (url) => {
  if (!url.includes('orcid.org') && !url.includes('create-account')) {
    cleanup()
  }
}
```

davidsmejia (approved):
Looks good just the one comment

PR #382, opened 2020-09-03
Import without create

kurtwheeler:
## Issue Number

N/A, David needs it

The goal is to just return the material JSON after importing, rather than actually saving it. This does that, along with fixing some bugs I found along the way.

## Types of changes

## Functional tests

Unit tests! They test the flow for importing materials including POSTing the material and then setting the grant.

## Checklist


davidsmejia (approved):
LGTM!

PR #387, opened 2020-09-08
Allow create user through sign in

Philinphiladelphia:
## Issue Number

N/a

This change allows the user to create their account through the sign-in button. If the user doesn't already exist, the proper calls will be made to create them.

## Types of changes

## Checklist


client/src/pages/account/index.js
```diff
@@ -1,16 +1,65 @@
...
+    decodeURI(`http://${req.headers.host}${req.url}`)
   )
+
+  const initialProps = {}
+
+  initialProps.orcid = response.response.orcid
+  initialProps.accessToken = response.response.access_token
+  initialProps.refreshToken = response.response.refresh_token
+
+  return initialProps
```
davidsmejia, line 62:
```suggestion
  const orcidRequest = await api.user.getORCID(
    query.code,
    decodeURI(`http://${req.headers.host}${req.url}`)
  )

  if (orcidRequest.isOk) {
    const {
      orcid,
      access_token: accessToken,
      refresh_token: refreshToken
    } = orcidRequest.response

    return {
      orcid,
      accessToken,
      refreshToken
    }
  }
```

client/src/api/index.js
```diff
@@ -147,6 +148,7 @@ export default {
       return [tokenRequest, userRequest]
     },
     create: async (orcid, accessToken, refreshToken, email, grants) => {
+      console.log(orcid, accessToken, refreshToken, email, grants)
```
davidsmejia, line 151: I think we should start removing console.logs at this point - I'm going to be turning on the lint rule soon

davidsmejia (approved):
some small changes but looks good!

kurtwheeler (approved):
Backend stuff looks good to me :+1:

PR #389, opened 2020-09-08
move material import before router.urls

davidsmejia:
## Issue Number
N/A was still having issues posting to this endpoint, this seems to resolve the issue


kurtwheeler (approved):
LGTM!

Issue #374, opened 2020-09-02
Integrate notification overhaul into report-issue endpoint


kurtwheeler:
### Context

#373 doesn't create the notification because #346 hasn't landed

### Problem or idea

Generate a notification for the user that reports an issue.

### Solution or next step

Use the notification text Deepa provided on https://github.com/AlexsLemonade/resources-portal/issues/345 to do so.

PR #391, opened 2020-09-09
Adds MATERIAL_REQUEST_REQUESTER_ESCALATED notification to report-issue endpoint.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/resources-portal/issues/374

This creates a notification when a user reports a request to the grants team.

This also changes it so the report-issue endpoint requires a `material_request_id` since the notification text and the message text reference the material request the user is asking for help with.

I think @davidsmejia may have been hoping to use this for other issues as well? If that is the case, then I think we need another set of email/notification copy and to use that if `material_request_id` is null.

## Types of changes

## Functional tests

I added unit tests.

## Checklist


api/resources_portal/views/report_issue.py
```diff
@@ -8,44 +8,98 @@
 
 from resources_portal.config.logging import get_and_configure_logger
 from resources_portal.emailer import send_mail
+from resources_portal.models import MaterialRequest, Notification
 
 logger = get_and_configure_logger(__name__)
 
 EMAIL_SOURCE_TEMPLATE = "Resources Portal Mail Robot <no-reply@{}>"
```
davidsmejia, line 15: do we need to update this?

davidsmejia (approved):
Looks good, I feel like it would be easier to read if it were just `get_issue_report_email` that optionally appended the material request info if the kwarg was there but looks good!

Issue #372, opened 2020-09-02
Weekly notification digest


kurtwheeler:
### Context

#346 does not send the weekly digest of notifications

### Problem or idea

Create a cron job to send the digest email every week. Probably using a management command.

### Solution or next step

See the attached zip file from @dvenprasad for what the emails should include: [weekly_digest.zip](https://github.com/AlexsLemonade/resources-portal/files/5163072/weekly_digest.zip)

kurtwheeler:
@dvenprasad is there a plain text version of this?

dvenprasad:
``` 
<Full Name>,
You received <X> notifications for the week of <date>.
--

<notification date>
<notification email copy>
<cta>


<notification date>
<notification email copy>
<cta>


<notification date>
<notification email copy>
<cta>


<notification date>
<notification email copy>
<cta>


The CCRR Team
-----
You are receiving this email because you subscribed to receive notifications from CCRR portal. Manage Notifications
Alex's Lemonade Stand Foundation
111 Presidential Blvd, Suite 203, Bala Cynwyd,  PA 19004
```

<notification email copy> -> this is the body of the corresponding notification. 
I would add 2 line breaks between each notification to create visual break (alternate would be to use `----` but that becomes visually noisy fast.)

PR #396, opened 2020-09-10
Weekly notification digest

kurtwheeler:
## Issue Number

#372 

Compiles and sends the weekly notification digests.

## Types of changes

## Functional tests

I added a unit test.

## Checklist


api/resources_portal/management/commands/send_weekly_digest_emails.py
```diff
@@ -0,0 +1,103 @@
...
+def send_user_weekly_digest(user):
+    start_time = user.weekly_digest_last_sent
+    if not start_time:
+        start_time = timezone.now() - datetime.timedelta(days=7)
+
+    pretty_start_time = start_time.strftime("%Y-%m-%d")
+    notifications = user.notifications.filter(created_at__gte=start_time)
+
+    if notifications.count() < 1:
+        print(f"No notifications for user: {user.full_name} this week!")
```
Philinphiladelphia, line 63:
```suggestion
        logger.info(f"No notifications for user: {user.full_name} this week!")
```

api/resources_portal/management/commands/send_weekly_digest_emails.py
```diff
@@ -0,0 +1,103 @@
...
+def get_plain_text_for_notification(notification):
+    email_dict = notification.get_email_dict()
+    plain_text_email_body = email_dict["plain_text_email_body"]
+
+    return f"{notification.human_readable_date}\n{plain_text_email_body}\n"
+
+
+def get_formatted_text_for_notification(notification):
+    email_dict = notification.get_email_dict()
+    pretty_time = utils.pretty_date(notification.created_at)
```
Philinphiladelphia, line 40: Have you checked that this outputs times in the user's timezone and not UTC? Django claims to have timezone support, which stores times in UTC and translates to the user's timezone in forms and templates. This might already be working automagically, but I would check. Reference: https://docs.djangoproject.com/en/3.1/topics/i18n/timezones/

Philinphiladelphia (commented):
Looking really good!! One question about UTC times/timezones, and you should be good to go.

api/resources_portal/management/commands/send_weekly_digest_emails.py
kurtwheeler: So this actually conveniently sidesteps that issue because `pretty_time` returns things like "5 minutes ago" or "3 days ago". Since both the created_at and the `timezone.now()` function use django times, it should be consistent. Additionally when we say "For the week of" we only specify down to the day, so that should be timezone safe as well.

This is actually really really convenient because we don't store what timezone each user is in.

Philinphiladelphia: Ah delightful. Good to go!

Philinphiladelphia (approved):
LGTM!!

PR #405, opened 2020-09-14
updating helpers

davidsmejia:
## Issue Number

Updates the helpers


Philinphiladelphia (approved):
Looks good. Might be better if the client was able to draw its test data from the JSON in the API dev_data, but having two sources is fine for now.

PR #411, opened 2020-09-15
update material schema

davidsmejia:
## Issue Number

Updates the material schema and implements a mechanism for custom error messaging


Philinphiladelphia (approved):
Looks good. Maybe there's some way to use the sequence map form to validate each member of the sequence_maps array? You've probably already considered this, I'm just musing on it myself.

PR #413, opened 2020-09-15
remove undefined env var

davidsmejia:
## Issue Number

The config file should be referencing the defined `isProduction` variable not the process.env directly


kurtwheeler (approved):
Cool, LGTM!

Issue #354, opened 2020-08-31
Integrate /notificatons


davidsmejia:
Needs to read from the api
Needs to update the header notifications badge based on localstorage last time checked, when not set should be able to have at least 1 to set value

PR #420, opened 2020-09-17
Integrate notifications

Philinphiladelphia:
## Issue Number

#354 

This adds notifications in the header, and bubbles on the account page/in the notifications tab of account. It handles updating the last time the user viewed notifications. It also adds support for links to requests, resources, and organizations in the notification.

## Types of changes

## Functional tests

None

## Checklist


api/resources_portal/models/frontend_notifications_config.py
```diff
@@ -0,0 +1,141 @@
...
+            "{other_name} linked a new grant {grant_name} with [organization_name]()."
+            "\nTeam members can now add resources associated with the grant."
+        )
+    },
+    "ORGANIZATION_INVITE": {
+        "body": "{organization_owner} has added you to their team, [organization_name]().",
+    },
+    "REPORT_TO_GRANTS_TEAM": {
+        "body": (
+            "Your issue has been reported to the ALSF Grants Team.<br>"
```
kurtwheeler, line 136: Some of these are using `<br>` and some are using `\n`. It should probably be consistent right? (I wonder if that was also an issue in the old config as well)

client/src/api/index.js
```diff
@@ -205,6 +205,12 @@ export default {
           method: 'PATCH',
           body: JSON.stringify(changes)
         })
+    },
+    notifications: {
+      list: (userId, authorization) =>
+        request(getAPIURL(`users/${userId}/notifications`), {
```
davidsmejia, line 211: adding a trailing slash will prevent an extra redirect

client/src/components/AccountLayout.js
```diff
@@ -1,38 +1,56 @@
...
   const router = useRouter()
+
+  let unreadNotifs = []
+  unreadNotifs = getUnreadNotifications()
   const links = [
     {
       text: 'Basic Information',
       href: '/account/basic-information',
-      notifications: 0
+      notifications: getNotificationsForType(unreadNotifs, 'basicInfo').length
```
davidsmejia, line 32: we should remove this for everything but the `notifications` sections

client/src/components/AccountLayout.js
```diff
@@ -1,38 +1,56 @@
 import React from 'react'
 import { Box, Heading, Main } from 'grommet'
 import { useRouter } from 'next/router'
+import { useNotifications } from 'hooks/useNotifications'
+import { NOTIFICATION_TYPES } from 'helpers/notificationTypes'
 import { useIsClient } from '../hooks/useIsClient'
 import { SideNav } from './SideNav'
 import Header from './Header'
 
+const getNotificationsForType = (notifications, type) => {
```
davidsmejia, line 10: you can set `notifications = []` so you dont have to bail...

If you set the default value then you can `return notifications.filter...` it will also be the empty array if you dont pass any to the function

kurtwheeler (approved):
The python parts of this LGTM, except for the one comment I made!

client/src/components/Notification.js
```diff
@@ -12,22 +12,89 @@ const readableTimeAgo = (time) => {
...
+      <Link href={href} key={1}>
+        <Anchor href="#" label={label || 'request'} />
+      </Link>
+    )
+  }
+  if (linkType === 'material_name') {
+    const href = `/resources/${notification.material.id}`
+    return (
+      <Link href={href} key={2}>
+        <Anchor href="#" label={label || notification.material.title} />
```
davidsmejia, line 28: can we set the href of the Anchor to match the link so if a user wants to open in a new tab via their browser it will work?

api/resources_portal/models/frontend_notifications_config.py
Philinphiladelphia: You're right. For this purpose, I changed it to all \n because we don't interpret this as HTML.

client/src/api/index.js
Philinphiladelphia: Added!

client/src/components/AccountLayout.js
Philinphiladelphia: Gotcha, removed

client/src/components/AccountLayout.js
Philinphiladelphia: This function was only needed for sorting notifications by tab, so it got removed entirely.

client/src/components/Notification.js
Philinphiladelphia: Yep! Done

client/src/pages/account/notifications/index.js
```diff
@@ -1,9 +1,34 @@
...
-  const notifications = []
+  const { notifications } = useNotifications()
+  const [didUpdateNotifs, setdidUpdateNotifs] = React.useState(false)
+  const { refreshUserData, user, token } = useUser()
+
+  React.useEffect(() => {
+    if (!didUpdateNotifs) {
+      setdidUpdateNotifs(true)
+      const updateNotifsViewed = async () => {
+        const today = new Date()
```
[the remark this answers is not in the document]
Philinphiladelphia: Done!

Issue #371, opened 2020-09-02
Outstanding notifications to create and/or test for


kurtwheeler:
### Context

#346 got way too big. I had to PR it. However some notification types I added didn't get used. Others didn't get tested.

### Problem or idea

These notification types should all be used at the appropriate places. Where is appropriate can be determined by looking at the notification text in the config.

### Solution or next step

The following notification types should be used wherever appropriate:
- [x] MATERIAL_REQUEST_SHARER_RECEIVED_INFO
- [x] MATERIAL_REQUEST_SHARER_ASSIGNMENT
- [x] MATERIAL_REQUEST_SHARER_EXECUTED_MTA
- [x] MATERIAL_ADDED
- [x] MATERIAL_ARCHIVED
- [x] MATERIAL_DELETED
- [x] ORGANIZATION_BECAME_OWNER
- [x] ORGANIZATION_NEW_OWNER
- [x] ORGANIZATION_NEW_GRANT
 
Additionally, these are used but not tested for:
- [x] MATERIAL_REQUEST_SHARER_REJECTED
- [x] MATERIAL_REQUEST_SHARER_CANCELLED
- [x] MATERIAL_REQUEST_REQUESTER_IN_FULFILLMENT
- [x] MATERIAL_REQUEST_REQUESTER_REJECTED
- [x] MATERIAL_REQUEST_REQUESTER_CANCELLED

kurtwheeler:
I took care of this a while ago but never closed it!

PR #421, opened 2020-09-17
Trigger and/or test outstanding notifications

kurtwheeler:
## Issue Number

#371 

This triggers the remaining notifications that I didn't cover before. It also tests a few that got missed.

I also added some date related fields to material_request just because I didn't wanna do a separate PR.

I also noticed that we never protected materials from deletes once they had requests so I added that while testing the notification for it.

## Types of changes

## Functional tests

Unit tests!

## Checklist


api/resources_portal/test/views/test_material.py
```diff
@@ -215,6 +224,17 @@ def test_delete_request_deletes_a_material(self):
...
+        self.assertEqual(
+            len(Notification.objects.filter(notification_type="MATERIAL_DELETED")),
+            self.organization.members.count(),
+        )
+
+    def test_delete_material_with_request_fails(self):
+        MaterialRequestFactory(material=self.material)
+        self.client.force_authenticate(user=self.user)
+        response = self.client.delete(self.url)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
Philinphiladelphia, line 236: I would consider making this error a 403? 400 gives the impress that the actual syntax of the request is wrong.

api/resources_portal/views/material_request.py
```diff
@@ -145,21 +142,25 @@ def has_object_permission(self, request, view, obj):
...
                 "material",
-                "created_at",
-                "updated_at",
             ]
 
         for field in forbidden_fields:
             if field in request.data:
                 attribute = getattr(obj, field)
                 if isinstance(attribute, models.Model):
                     # UUID's can't just be treated as strings for some reason...
```
Philinphiladelphia, line 154: lmao this represented 90% of the issues I ran into when I first started working here. You know my pain

Philinphiladelphia (approved):
Looks good! At some point I would consider checking emails whenever we check number of notifications sent, but that doesn't need to happen right now... right?

api/resources_portal/test/views/test_material.py
kurtwheeler: Hmm, not sure I agree. 

> 400 Bad Request response status code indicates that the server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400

I would consider this to fall under the category of `invalid request message framing`. Basically the request is invalid because you cannot delete a material with requests. I think a 403 might imply that someone was allowed to? (I don't think a 403 would be wrong here necessarily, but IMO 400 is better.)

kurtwheeler:
> Looks good! At some point I would consider checking emails whenever we check number of notifications sent, but that doesn't need to happen right now... right?

Definitely not now... I don't know about "whenever we check the number of notifications sent", but it might be good to have one or two tests test that email only gets set for notifications the user has opted in for.

PR #424, opened 2020-09-17
Squash migrations down to one bigun.

kurtwheeler:
## Issue Number

N/A, been wanting to do this before beginning testing on staging.

Squashes all the migrations. Before this is deployed, I'll ssh onto the server and clear out the database so it starts fresh. I want to do this anyway because we're about to be the test users rather than needing that all pre-populated.

## Types of changes

- Cleanup

## Functional tests

Unit tests will verify that I broke nothing.

## Checklist


Philinphiladelphia (approved):
LGTM!

PR #428, opened 2020-09-18
fixed grants and personal org bug

Philinphiladelphia:
## Issue Number

N/a

This fixes a bug where grants were not a list if there was only one of them, and also ensures that personal org is created. I checked the database, and personal org is indeed created. Note that I checked the "teams" page of accounts and the personal org still doesn't appear there- not sure whether it should or not.

## Types of changes

## Functional tests

Manual tests

## Checklist


davidsmejia:
That is correct the personal org shouldn't appear in the `teams` tab. I was seeing it null (in GET: `/users/[id]`) after creating the user with a single grant.

davidsmejia (approved):
LGTM

PR #430, opened 2020-09-18
Remove temporary code to reset staging database.

kurtwheeler:
## Issue Number

#424 

#424 added a change to taint the database to reset migrations. This undoes that so we don't lose our database every deploy.


Philinphiladelphia (approved):
Look good!

Issue #353, opened 2020-08-31
Integrate /account/requests


davidsmejia:
/listing page is done
/detail page needs to be integrated

PR #432, opened 2020-09-18
request management

davidsmejia:
## Issue Number

#353 

Implements /account/requests and integrates it. Also fixes some bugs I found along the way.

TODO:
Move the management api logic into a hook that the forms use.
Break the steps into their own components, each one isnt that big but all together is too much.

## Types of changes

## Functional tests

Created resource
Set all requirements
Was able to go through each step and mark fulfilled

## Checklist

## Screenshots

TBD

### DIFF
--- a/client/src/api/index.js
+++ b/client/src/api/index.js
@@ -126,11 +126,16 @@ export default {
       }),
     requests: {
       filter: (id, query = {}, authorization) =>
-        request(getAPIURL(`materials/${id}/requests/`, query), {
+        request(getAPIURL(`materials/${id}/requests`, query), {
           authorization,
           method: 'GET'
         })
-    }
+    },
+    delete: (id, authorization) =>
+      request(getAPIURL(`materials/${id}/`), {
+        authorization,
+        method: 'DELETE'
+      })
   },
   user: {
     get: userGetInfo,
@@ -350,17 +355,21 @@ export default {
       body: JSON.stringify({ email })
     }),
   requests: {
+    get: (materialRequestId, authorization) =>
+      request(getAPIURL(`material-requests/${materialRequestId}/`), {
+        authorization
+      }),
     create: (materialRequest, authorization) =>
       request(getAPIURL('material-requests/'), {
         authorization,
         method: 'POST',
         body: JSON.stringify(materialRequest)
       }),
-    update: (materialRequestId, materialRequest, authorization) =>
+    update: (materialRequestId, changes, authorization) =>
       request(getAPIURL(`material-requests/${materialRequestId}/`), {
         authorization,
         method: 'PATCH',
-        body: JSON.stringify(materialRequest)
+        body: JSON.stringify(changes)
       }),
     list: (authorization) =>
       request(getAPIURL('material-requests/'), {
@@ -369,6 +378,19 @@ export default {
     filter: (filter, authorization) =>
       request(getAPIURL('material-requests/', filter), {
         authorization
-      })
+      }),
+    notes: {
+      add: (materialRequestId, text, authorization) =>
+        request(
+          getAPIURL(
+            `material-requests/${materialRequestId}/fulfillment-notes/`
+          ),
+          {
+            authorization,
+            method: 'POST',
+            body: JSON.stringify({ text, material_request: materialRequestId })
+          }
+        )
+    }
   }
 }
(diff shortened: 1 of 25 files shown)


client/src/api/index.js
```diff
@@ -126,11 +126,16 @@ export default {
       }),
     requests: {
       filter: (id, query = {}, authorization) =>
-        request(getAPIURL(`materials/${id}/requests/`, query), {
+        request(getAPIURL(`materials/${id}/requests`, query), {
```
Philinphiladelphia, line 129: Trailing / on this one is not neccessary?

davidsmejia: it actually doesn't work at the moment

client/src/components/ManageRequestCard.js
```diff
@@ -1,70 +1,146 @@
...
+  const isRequester = isResourceRequester(request)
+
+  const showSharerWaiting =
+    !isRequester && config.status === 'AWAITING_ADDITIONAL_DOCUMENTS'
+  const showRequesterAction =
+    isRequester && config.status === 'AWAITING_ADDITIONAL_DOCUMENTS'
+  const showActionPill = showSharerWaiting || showRequesterAction
+  const pillLabel = showSharerWaiting
+    ? 'Waiting on Requester'
+    : 'Reuires Action'
```
Philinphiladelphia, line 99:
```suggestion
    : 'Requires Action'
```

client/src/components/ManageRequestsTable.js
```diff
@@ -1,20 +1,184 @@
...
+      (r) => !assignedTo || r.assigned_to === assignedTo
+    )
+
+    const filteredTeam = filteredAssignedTo.filter(
+      (r) => !team || r.material.organization.id === team
+    )
+
+    const filteredRequests = [...filteredTeam]
+
+    const cleanedFilteredRequests = unique(filteredRequests, 'id')
```
Philinphiladelphia, line 100: Should the sorting to make sure that they're unique happen before you filter them on all the other attributes?

davidsmejia: It actually won't really affect the filtered requests here since they are essentially chained. This was more for if we wanted to do OR logic instead of AND which would be if we filtered each one from the original list and then merged that list. But also depending on the source of the requests it is an additional check that has not side effect other than the list will not have dupes

client/src/components/resources/RequesterManageRequestForm.js
```diff
@@ -0,0 +1,365 @@
...
+              <PreviewAbstract abstract={request.requester_abstract} />
+            )}
+            {request.address && <PreviewAddress address={request.address} />}
+          </Box>
+        )}
+        {state === 'AWAITING_ADDITIONAL_DOCUMENTS' && (
+          <>
+            <Box pad={{ bottom: 'large' }}>
+              <Text>
+                You're request has been accepted on the conditional that you
```
Philinphiladelphia, line 186:
```suggestion
                Your request has been accepted on the condition that you
```

client/src/hooks/useUser.js
```diff
@@ -47,6 +47,50 @@ export const useUser = (defaultUser, defaultToken) => {
...
+  const isPersonalResource = (resource) => {
+    if (typeof resource.organization === 'number') {
+      return resource.organization === user.personal_organization.id
+    }
+    if (typeof resource.organization.id === 'object') {
+      return resource.organization.id === user.personal_organization.id
+    }
+
+    // there was no way to tell
+    // since no org was defined
```
Philinphiladelphia, line 74: Is this referring to the personal org not being created? That should be fixed now.

client/src/pages/account/requests/index.js
```diff
@@ -1,14 +1,51 @@
...
+  const [received, setReceived] = React.useState([])
+  const [sent, setSent] = React.useState([])
+  const hasReceived = received && received.length > 0
+  const hasSent = sent && sent.length > 0
+  const hasAnyRequests = hasReceived || hasSent
 
-  const hasAnyRequests = receivedRequests.length > 0 || sentRequests.length > 0
+  React.useEffect(() => {
+    const fetchRequests = async () => {
+      const requestsRequest = await api.requests.filter({ limit: 1000 }, token)
```
Philinphiladelphia, line 25: We should consider making the api endpoint filterable. not a concern for now, but soon.

Philinphiladelphia (approved):
Looks good! Couple of typos and some stubs, but definitely good enough to work with.

client/src/pages/account/requests/index.js
davidsmejia: Yeah I forget why we moved away from that but we did discuss it.

client/src/hooks/useUser.js
davidsmejia: it should always work but when sentry is added this should be sent there

PR #433, opened 2020-09-18
adding search empty and fixing paths

davidsmejia:
## Issue Number

Handles the search empty and no results states. Also uses the newer absolute import paths.


Philinphiladelphia (approved):
LGTM 🚀

PR #435, opened 2020-09-18
send to api on submit

davidsmejia:
## Issue Number

Refreshes the user on basic-account load and also sends the new email to the api and refreshes the user on submit


Philinphiladelphia (approved):
LGTM

PR #436, opened 2020-09-18
fix links on next-steps

davidsmejia:
N/A

Send the new user to the correct locations.
Also sets the hrefs on the anchors


Philinphiladelphia (approved):
LGTM

PR #423, opened 2020-09-17
Update the elasticsearch index every five minutes.

kurtwheeler:
## Issue Number

N/A We need to update the elasticsearch index periodically so new resources show up in the search.

## Purpose/Implementation Notes

This uses a cron job on the API to do so.

This is entirely based off of what refinebio does, and it seems to work there pretty well.

## Types of changes

- New feature (non-breaking change which adds functionality)

## Functional tests

None until it deploys!

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules

PR #437, opened 2020-09-21
Found another place we needed to change from refinebio!

kurtwheeler:
## Issue Number

#423 

The cron log is full of 
```
Error: No such container: dr_api
```
so at least it's running frequently.

## Types of changes

## Functional tests

None till staging!


davidsmejia (approved):
ooph

Issue #422, opened 2020-09-17
Stop logging cron output


kurtwheeler:
### Context

I'm making a cron job to update the ES index.

### Problem or idea

I'm currently logging the output in case I need to debug it.

### Solution or next step

We should stop logging that once we know it works.

kurtwheeler:
Closed by #438

PR #438, opened 2020-09-21
Only log the most recent cron output so we don't fill our disk.

kurtwheeler:
## Issue Number

#422 
#423 

It's working!!! Now we don't need all the logs. If something is going wrong we can look at the most recent and get an idea of what that is.

## Types of changes


davidsmejia (approved):
👍

PR #439, opened 2020-09-22
Address line 2 shouldn't be required!

kurtwheeler:
## Issue Number

N/A just noticed this while requesting a resource.

Address line 2 shouldn't be required.


davidsmejia (approved):
good catch!

PR #444, opened 2020-09-24
Fix some broken html, regenerate inlined HTML.

kurtwheeler:
## Issue Number

#335

One is still small and off center. Looks like a closing tag issue, so I've fixed a couple html things and regenerated the inlined html.

## Types of changes


davidsmejia (approved):
LGTM maybe add a title or remove it

PR #445, opened 2020-09-24
I think there was some drift. This attempts to reconcile that.

kurtwheeler:
## Issue Number

#444 

## Purpose/Implementation Notes

I guess I had changed more than I thought.

## Types of changes

- Bugfix (non-breaking change which fixes an issue)

PR #446, opened 2020-09-24
Update inlined email because that's actually what gets used.

kurtwheeler:
## Issue Number

#445 

445 did nothing, this is what actually needed to change.


davidsmejia (approved):
Lgtm

PR #447, opened 2020-09-24
The non-inlined had apparently fallen way far behind.

kurtwheeler:
## Issue Number

#446 

I think this should be about what we've been needing. I think.


davidsmejia (approved):
🚀

PR #448, opened 2020-09-24
Tried another inlining tool.

kurtwheeler:
## Issue Number

#447 

Now the email looks poorly formatted, so we're trying a new inliner.


davidsmejia (approved):
this looks better!

Issue #452, opened 2020-09-25
Have api change material request status to IN_FULFILLMENT when approved and requester uploads required docs


davidsmejia:
### Context

When an `APPROVED` material_request is updated by the requester and it will not require a full executed MTA from the sharer. (Meaning only an IRB or `payment_info`/`payment_method_notes` are requirements) We should kick the material_request to `IN_FULFILLMENT` in the same post that updates the material_request to have those new values.

This is to prevent inserting another "review" state into the material_request pipeline between `APPROVED` and `IN_FULFILLMENT`

### Solution or next step

When we send the notification for `MATERIAL_REQUEST_SHARER_RECEIVED_INFO` here:
https://github.com/AlexsLemonade/resources-portal/blob/dev/api/resources_portal/views/material_request.py#L401
We should also set the material_request to `IN_FULFILLMENT` 

Also add line to the `MATERIAL_REQUEST_SHARER_RECEIVED_INFO` notification that has:
```Please review the items and make arrangements to send the <resource type> to <Requester>.```
Also change the CTA from `Review Items` -> `View Request`

**Comment by davidsmejia** [comment]
@kurtwheeler when we talked about it last this was our approach. just more context to understand the original problem. It was from the 403 error when the client was posting IN_FULFILLMENT. 

```MANAGE REQUESTS
[ ] remove in_fulfillment if mta is not required line: 127 in requester form
[ ] show state that all documents are present check is_missing_requester_documents

if it doesn't require action sharer then it should be checked to be placed in_fulfillment
otherwise it requires action sharer and needs the MTA
ask deepa if we need another state when IRB is uploaded but the next step is only in_fulfillment
```

PR #453, opened 2020-09-25
Allow requesters to move requests to IN_FULFILLMENT if no MTA is required.

kurtwheeler:
## Issue Number

#452 

If a material does not need an MTA, then once a user uploads the IRB they should also be able to move the request to in fulfillment without needing additional approval. This makes the material request view allow that.

## Types of changes

## Functional tests

I've added 2 unit tests for this!

## Checklist


davidsmejia (approved):
LGTM!

PR #455, opened 2020-09-25
remove the host from the sequence map preview source

davidsmejia:
## Issue Number
N/A fixes an issue with sequence map image preview where it was expecting the host to be missing from the download_url


kurtwheeler (approved):
LGTM!

PR #466, opened 2020-09-29
remove bar from more options

davidsmejia:
## Issue Number

N/A there was some artifact in the `More Options` dropdown - there was extra text that was supposed to have been removed.
This removes it

Screen Shot 2020-09-29 at 12 38 40 PM


kurtwheeler (approved):
LGTM!

PR #488, opened 2020-10-06
User.addresses now filters out unsaved addresses.

kurtwheeler:
## Issue Number

#461 

Addresses that weren't saved for reuse were being returned as `None` instead of being filtered out.

## Types of changes

## Functional tests

There was a unit test that was testing that this wasn't working. When I fixed the issue the test broke and now I fixed the test.

## Checklist


davidsmejia (approved):
🚗 💨

Issue #486, opened 2020-10-06
Get sentry working


kurtwheeler:
### Problem or idea

We need sentry set up for both the frontend and backend.

### Solution or next step

Set it up.

kurtwheeler:
Once https://github.com/AlexsLemonade/resources-portal/pull/502 is merged this should be complete!

PR #497, opened 2020-10-07
client sentry

davidsmejia:
## Issue Number

#486 

This is the initial integration. 

We need to:
 * break the staging and prod apps on vercel into different projects and then only deploy to prod via github actions
 * also we need to move the env vars into the production env vars on now.sh


kurtwheeler (approved):
This looks like a good start!

PR #502, opened 2020-10-08
implement source maps for webpack build

davidsmejia:
## Issue Number

N/A

Adds `@zeit/next-source-maps` to the build process for better debugging with sentry


kurtwheeler (approved):
Oh wow, that setup looks so minimal!!!

Issue #503, opened 2020-10-08
Incorrect Grant info failed to submit


jashapiro:
### Context

When logging in for the first time, I decided to see what would happen if I clicked that the grant info was incorrect:

### Problem or idea

I entered some text, clicked send, and got an error.

<img width="1318" alt="Screen Shot 2020-10-08 at 1 43 54 PM" src="https://user-images.githubusercontent.com/20483/95494711-514c3980-096c-11eb-8ee6-83a9a1f45b96.png">

### Solution or next step

I would assume we should not get an error here, and this form should generate a ticket somewhere...

jaclyn-taroni:
Here's what happened when I clicked on `Report missing/incorrect information`: 

<img width="1507" alt="Screen Shot 2020-10-08 at 2 50 09 PM" src="https://user-images.githubusercontent.com/19534205/95500946-9a54bb80-0975-11eb-8be4-47a5c0f272e7.png">

I'm going to submit the information in a second, but wanted to record this screenshot.

jaclyn-taroni:
I got the same error banner as Josh after entering the text. Rereading the original post, I did not understand when the error occurred and it appears I encountered the same problem.

kurtwheeler:
It occurred for a different reason this time!!!

jaclyn-taroni:
For some additional context, I _believe_ I accidentally said that it was correct, hit the back button, and then I went through the steps to report incorrect information.

kurtwheeler:
@cansavvy @cbethell @envest if you haven't gone through the flow yet I'd be curious to know if it happens to you too.

kurtwheeler:
I just created a new account to test this. It works!

PR #504, opened 2020-10-08
Make embeded images default to empty list instead of None.

kurtwheeler:
## Issue Number

#503 

The error was:
```
2020-10-08T13:44:57.192-04:00 | for embedded_image in embedded_images:
2020-10-08T13:44:57.192-04:00 | TypeError: 'NoneType' object is not iterable

```

This doesn't get tested in unit tests because we can't actually send mail without being deployed.

## Types of changes

## Functional tests

I can't really without staging.


davidsmejia (approved):
LGTM

PR #510, opened 2020-10-08
next 9.4.2 + set webpack config devtool to source-map + productionBrowserSourceMaps

davidsmejia:
## Issue Number

N/A 

Ok so as it turns out nextjs was preventing the building of source maps for production on build but one minor version later they added an experimental config option to do just that. So this PR does just that.


cgreene (approved):
Let's get some easy to read error reports! 🚀

Issue #474, opened 2020-10-02
Don't let users import same resource again


dvenprasad:
### Context

Users can import the same resource more that once.

### Problem or idea

Don't let them do that. 

### Solution or next step

_You can tag others or simply leave it for further investigation, but you must propose a next step towards solving the issue._

PR #518, opened 2020-10-09
Block redundant imports

Philinphiladelphia:
## Issue Number

#474 

This prevents users from importing the same resource twice on the back-end. It also adds tests for importing.

## Types of changes

## Functional tests

Added a test suite

## Checklist

## Screenshots


api/resources_portal/test/views/importers/test_geo.py
```diff
@@ -76,10 +84,36 @@ def test_import_succeeds_for_study_without_pubmed_id(self):
...
+    @patch("resources_portal.importers.geo.gather_all_metadata", side_effect=get_mock_dataset_data)
+    def test_import_previously_imported_geo_fails(self, mock_dataset):
+        self.client.force_authenticate(user=self.user)
+        accession_code = "12345"
+        response = self.client.post(
+            self.url, {"import_source": "GEO", "accession_code": accession_code}
+        )
+        import pdb
+
+        pdb.set_trace()
```
kurtwheeler, line 100: PDB!

api/resources_portal/views/import_materials.py
```diff
@@ -109,10 +112,41 @@ def create(self, request, *args, **kwargs):
...
             )
         elif import_source == "PROTOCOLS_IO":
+            for material in Material.objects.filter(imported=True, import_source="PROTOCOLS_IO"):
+                if (
+                    "protocol_doi" in material.additional_metadata
+                    and request.data["protocol_doi"] == material.additional_metadata["protocol_doi"]
+                ):
+                    return JsonResponse(
+                        {
+                            "error": f'A material with doi {request.data["protocol_doi"]} has already been imported.',
```
kurtwheeler, line 144:
```suggestion
                            "error": f'A material with DOI {request.data["protocol_doi"]} has already been imported.',
```

kurtwheeler (changes requested):
This looks good for the most part. The biggest issue I see is that `pdb.set_trace()` should be failing the tests and its not. I think we might need an `__init__.py` in the `importers` directory.

Other than that this looks good. Will the `error` of the response be displayed to the user in the frontend? Just wondering how this will end up looking to users.

kurtwheeler (changes requested):
Ok so I was just chatting with David a bit and he pointed out that rather than preventing the `import` endpoint from working, we should instead move the logic to the material endpoint and prevent the material from being created twice. He also pointed out that if there is an error we should return the material that already exists so the user can see that it was imported and who did it.

kurtwheeler (approved):
This LGTM!!

PR #519, opened 2020-10-09
if user id isnt set logout user

davidsmejia:
## Issue Number

n/a 

I'm seeing issues in sentry that user.id isnt defined. this should help recover browsers that have bad localstorage issues.


kurtwheeler (approved):
This makes sense to me!

Issue #511, opened 2020-10-08
Creating team and adding members 


envest:
### Context

I signed in and my first instruction was to add members to a team. I clicked a link which took me to this page:

<img width="1438" alt="Screen Shot 2020-10-08 at 4 35 20 PM" src="https://user-images.githubusercontent.com/10382980/95513022-a6e20f80-0987-11eb-99c8-dcb76eb16fd9.png">

### Problem or idea

Nothing shows up for me on this page. Just the wobbly lemon (which I love)

### Solution or next step

What I imagine exists here is... a table of existing teams I am part of and my role and a button that says "create team". Each existing team in the table might have a button that says "add members".

kurtwheeler:
I think this may have been an intermittent error (we're working on a good resolution for that).

Can you give this another shot? The page seems to be loading for me.

envest:
Thanks... I got the same non-loading event today multiple times :/

My steps:
1. Navigate to https://staging.resources.alexslemonade.org/create-account?grant_title=Steven%27s%20Grant&grant_funder_id=ALSF&grant_title=Steven%27s%20Other%20Grant&grant_funder_id=ALSF
2. Click "Sign in with ORCID ID"
3. Enter my real email (stevenmasonfoltz@gmail.com)
4. Verify grant information ("This information is correct")
5. On Next Steps, I click "Add members of your lab..."
6. My Account > Team page shows a lemon
7. My Account > Notifications looks like I expect
8. My Account > Manage Resources == has a header and a lemon
9. My Account > Requests == just a lemon
10. My Account > Basic Information == "Uh oh, something happened" + dialog box for reporting an issue

envest:
This is resolved from my perspective

kurtwheeler:
Sweet I think it can be closed then!

PR #520, opened 2020-10-09
Apparently family name can be null in orcid.

kurtwheeler:
## Issue Number

#511 

I'm seeing errors like:
```
2020-10-09T11:29:16.186-04:00 | File "./resources_portal/views/user.py", line 189, in create
2020-10-09T11:29:16.186-04:00 | last_name = summary["person"]["name"]["family-name"]["value"]
2020-10-09T11:29:16.186-04:00 | TypeError: 'NoneType' object is not subscriptable
```

I guess we made an assumption that people would have family names in orcid, but that's not always true apparently.

## Types of changes

## Functional tests

Working on that now!

### DIFF
--- a/api/resources_portal/views/user.py
+++ b/api/resources_portal/views/user.py
@@ -186,7 +186,8 @@ class UserViewSet(viewsets.ModelViewSet):
         # Get first and last name
 
         first_name = summary["person"]["name"]["given-names"]["value"]
-        if "family-name" in summary["person"]["name"]:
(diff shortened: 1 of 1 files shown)


davidsmejia (approved):
👪 📛

kurtwheeler:
I'm setting family name to blank instead of overriding the model. I feel like this isn't ideal but this way we don't have to worry about things that assume that field isn't null.

Issue #493, opened 2020-10-07
PDX Form Field issues


dvenprasad:
### Problem or idea
- Section labels are not formatted correctly
- Some field labels are not formatted correctly
   - histology
  - tumor_chracterization_technology
   - is_not_of_EBV_origin
  - is_passage_QA_performed
  - response_to_QA_performed
  - tumor_omics
  - matastases_in_strain 
- Optional fields are not marked

### Solution or next step

_You can tag others or simply leave it for further investigation, but you must propose a next step towards solving the issue._

PR #543, opened 2020-10-16
fix pdx form fields

davidsmejia:
## Issue Number

#493 

This should fix all pdx created from now on. We could write a migration to fix the existing PDX's but I think it would be easier to trash the staging DB and do testing again in the future. I decided that all lower case would be best for all attributes in additional_metadata.

## Types of changes

## Functional tests

N/A

## Checklist

## Screenshots

n/a


Philinphiladelphia (approved):
Looks good.

Issue #526, opened 2020-10-09
Notification emails are not displaying properly in Apple Mail client


jaclyn-taroni:
### Problem or idea

For both notification emails I received the contents displayed in the Apple mail client are: 

<img width="893" alt="Screen Shot 2020-10-09 at 12 28 10 PM" src="https://user-images.githubusercontent.com/19534205/95617508-8d010500-0a39-11eb-97c3-742094104f8f.png">

@jashapiro reported the same problem. Navigating to Gmail in the browser I see:

<img width="972" alt="Screen Shot 2020-10-09 at 2 15 07 PM" src="https://user-images.githubusercontent.com/19534205/95617775-f2ed8c80-0a39-11eb-9348-421547c2515f.png">

Notably this added to my confusion reported in #522.

### Solution or next step

_You can tag others or simply leave it for further investigation, but you must propose a next step towards solving the issue._

jashapiro:
I would add that the links to "Manage Notifications" and "Terms of Use" did not work. Also, I feel like even the full email is lacking in context. We should maybe have a bit more copy in there.

jaclyn-taroni:
I do not have this problem with refine.bio emails if that is helpful information.

dvenprasad:
What other information would you like to see here? @jashapiro

jashapiro:
I'm not sure... something about what I can do now that I am on a team? What does this mean for me? And maybe "View Team" should say something like "View Team on Resources Portal"? Something to give more meat to the email.

Also, I seem to recall that only people who have already signed up for the portal can be added? If that is true, then we don't need "what is the resources portal", but I feel like that is something we will need to add later.

kurtwheeler:
I've fixed this specific issue. There's still #595, but that's a much more minor bug that doesn't render the emails unusable by a portion of our users.

PR #554, opened 2020-10-20
Add a tag to disable apple's email reformatting.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/resources-portal/issues/526

Not sure if this'll do it or not... But it seems to be worth a shot based on https://stackoverflow.com/questions/46543783/apple-mail-ios-11-template-sizing and the fact that Apple mail seems to be the only client that we've gotten an issue reported for yet, and there's a thing to disable their reformatting.

## Types of changes

## Functional tests

Gotta deploy to staging for this.


davidsmejia (approved):
LGTM

Issue #556, opened 2020-10-20
Cannot delete or archive materials


kurtwheeler:
### Context

I was trying to archive or delete a material to generate a notification.

![Screenshot_2020-10-20_16-28-48](https://user-images.githubusercontent.com/4026833/96640706-8dc75000-12f1-11eb-9671-e3dec30e9e3b.png)

### Problem or idea

It shows a modal asking me if I'm sure, then when I click yes nothing happens. For both archiving and deleting.

### Solution or next step

Fix them!

kurtwheeler:
The manage resource card needs to call API methods on `resources` not `resource`.

PR #575, opened 2020-10-27
556 archive and delete resource from dropdown button then modal

davidsmejia:
## Issue Number

#556 

I thought this issue was limited to using the incorrect api function.
turns out the modal was being dismissed before the button was being pressed because the modal was nested inside of the dropdown button. This PR moves that outside and toggles the button dropdown manually after a modal link is clicked.

## Types of changes

## Functional tests

n/a tested archiving and deleting resources

## Checklist

## Screenshots

n/a


davidsmejia:
The issue isnt really blocked but testing will be a sane process when that is pushed.

kurtwheeler (approved):
LGTM!

PR #581, opened 2020-10-27
return valid if not an import source we validate for uniqueness

davidsmejia:
## Issue Number

N/A

Was getting an error when trying to import from a non-automagical import source


davidsmejia:
closed in favor of #582

PR #582, opened 2020-10-27
Fix importing other resource types, adds test, adds fields to MaterialRelationSerializer

### LINKED ISSUE #576: Add is_archived to /organisation/[id]/materials response
### Context

[#575](https://github.com/AlexsLemonade/resources-portal/pull/575) wont really work because since the attribute is not present in the response. 

To be clear the client updates the material correctly but you can't tell because it only shows `Publicly Available` when is_archived is false and since it is undefined it always shows up as publicly available even though the resource is in fact archived.

### Solution or next step

Add is_archived to the serializer

kurtwheeler:
## Issue Number

#576 

Fixes #576 and another bug along the way with importing other resource types.

## Types of changes

## Functional tests

Unit tests!!!

## Checklist


davidsmejia (approved):
lgtm

PR #584, opened 2020-10-27
Try attachment instead of inline.

kurtwheeler:
## Issue Number

#583 

Attachments might work better than inline.

## Types of changes


davidsmejia (approved):
LGTM

PR #585, opened 2020-10-27
adding images

davidsmejia:
## Issue Number

N/a

this will make the images used in the notification emails available at `staging.resource.alexslemonade.org/alexs-logo.png`

Same with prod when it is merged


kurtwheeler (approved):
LGTM! I should be able to take it from here.

PR #588, opened 2020-10-27
Remove logo files, instead link to images in emails.

kurtwheeler:
## Issue Number

#585 

This uses the links to images that were added so the emails don't need to embed the images because Apple mail doesn't support that.

## Types of changes

## Functional tests

Wanna test on staging.


davidsmejia (approved):
LGTM I think some of the space needs to be adjusted again tho

Issue #571, opened 2020-10-26
Management Commands For re-naming material.additional_metadata attributes


davidsmejia:
### Context

Since the client is responsible for defining additional_metadata keys we need to be able to migrate the keys in that object. 

### Problem or idea

Create a management command for migrating additional_metadata keys.
The command should:
 * take two arguments old_key_name and new_key_name
 * query the db for models that contain that key in their additional_metadata attribute
 * change the key name to the new name with the same value
 * save the model

Pagination might be something worth considering.

### Solution or next step

Create the command

davidsmejia:
Would be good for the command to also take a material category filter and maybe also support adding and removing which changing keys would use internally.

PR #636, opened 2020-11-13
Material metadata migrations

Philinphiladelphia:
## Issue Number

#571 

This adds a management command that allows adding, removing, and changing metadata keys in additional_metadata for all materials in the database or all materials of a given category.

## Types of changes

## Functional tests

Only manual tests

## Checklist

## Screenshots


api/resources_portal/management/commands/material_additional_metadata_key_change.py
```diff
@@ -0,0 +1,119 @@
...
+        if not validateCategory(category):
+            logger.error(f"The provided category '{category}' is not a valid material category.")
+            return
+        querySet = Material.objects.filter(category=category)
+    else:
+        querySet = Material.objects.all()
+
+    with transaction.atomic():
+        for material in querySet:
+            material.additional_metadata[new_key] = ""
```
kurtwheeler, line 28: It seems like the default value should be a parameter right?

api/resources_portal/management/commands/material_additional_metadata_key_change.py
```diff
@@ -0,0 +1,119 @@
...
+
+    if category:
+        if not validateCategory(category):
+            logger.error(f"The provided category '{category}' is not a valid material category.")
+            return
+        querySet = Material.objects.filter(category=category)
+    else:
+        querySet = Material.objects.all()
+
+    with transaction.atomic():
```
kurtwheeler, line 26: I'm not sure we want to do this in a transaction. It'd be fine for now, but as we scale that could end up being a problem. Especially because this function is idempotent, so even if it is interrupted and has to be restarted it'd be fine right?

api/resources_portal/management/commands/material_additional_metadata_key_change.py
```diff
@@ -0,0 +1,119 @@
...
+            return
+        querySet = Material.objects.filter(additional_metadata__has_key=old_key, category=category)
+    else:
+        querySet = Material.objects.filter(additional_metadata__has_key=old_key)
+
+    # Note that this doesn't use the above add and remove commands. This is because doing so would require twice the queries.
+    with transaction.atomic():
+        for material in querySet:
+            material.additional_metadata[new_key] = material.additional_metadata[old_key]
+            material.additional_metadata.pop(old_key)
```
kurtwheeler, line 72:
```suggestion
            material.additional_metadata[new_key] = material.additional_metadata.pop(old_key)
```

api/resources_portal/management/commands/material_additional_metadata_key_change.py
```diff
@@ -0,0 +1,119 @@
...
+
+# This is a command to add, remove, or change the names of additional metadata fields currently in the server.
+# It has the following use cases:
+# 1. Change a metadata key
+# ./rportal migrate-metadata-keys old_key new_key --category [CATEGORY]
+# 2. Add a metadata key
+# ./rportal migrate-metadata-keys --add new_key --category [CATEGORY]
+# 2. Remove a metadata key
+# ./rportal migrate-metadata-keys --remove old_key --category [CATEGORY]
+# For each use case, the category parameter is optional, and categories should be specified in all caps, like "PLASMID" or "MODEL_ORGANISM".
```
kurtwheeler, line 87: This comment would make a great docstring for the class.

kurtwheeler (commented):
This looks good to me aside from a few minor comments. I'll let @davidsmejia approve it though!

api/resources_portal/management/commands/material_additional_metadata_key_change.py
Philinphiladelphia: David had mentioned doing this in a paginated fashion. I can go ahead and do so, though it'll be a little difficult

api/resources_portal/management/commands/material_additional_metadata_key_change.py
Philinphiladelphia: good catch

api/resources_portal/management/commands/material_additional_metadata_key_change.py
davidsmejia: Yeah you should just operate on slices of the query set at a time modify that slice then write that slice (page) to the db

api/resources_portal/management/commands/material_additional_metadata_key_change.py
Philinphiladelphia: Good point. Done!

api/resources_portal/management/commands/material_additional_metadata_key_change.py
kurtwheeler: Just make sure that if you don't call `save()` on each object that you update `updated_at` manually because bulk updates won't do it.

api/resources_portal/management/commands/material_additional_metadata_key_change.py
Philinphiladelphia: Just added it!

davidsmejia:
This is looking good, 

1)
I would note that due to the nature of why we need this command I would probably add some checks to determine if a key exists and act appropriately. (ie if there is a bug on the client that is writing a bad key to the metadata, since it isnt regulated in a meaningful way on the client (it is trivially) it would be safer to be able attempt to create/change/remove the keys but also return some statistics or model ID's that show if the operation wasn't possible so we can try to rectify that after the command takes care of the majority of changes requested.

2) You could pull out the selecting of categories into it's own shared function and throw an error if the category doesnt exist. Since this will be run as an administration command it would be like trying to ssh to a server with a bad key i would expect that command to exit non 0

3) I would consider instead of transactions doing something like:
Have the above function return a queryset then operate on that at 100 or so at a time,  iterate to make your changes, then bulk update and make sure updated_at is set to now()

Bulk Update docs here:
https://docs.djangoproject.com/en/3.1/ref/models/querysets/#bulk-update

@kurtwheeler can i get your thoughts on this too?

kurtwheeler:
Those sound like good suggestions @davidsmejia . It didn't occur to me that this would end up failing if the dict didn't have a key that was being removed.

davidsmejia:
This is what I was thinking just in terms of layout not sure if the logic holds up.

```psuedocode


def action_method_example(materials):
   // change each material
   // return changes materials
  return materials


def Command:
   // get queryset or throw if bad category
  materials = get_materils(category=ALL)
  // paginate
  page = 0
  pagesize = 100
  for materials in queryset[page:pagesize*(page+1)]:
     modified_materials = action_method_example(materials)
     Material.objects.bulk_update(modified_materials, ['additional_metadate', 'updated_at'])

```

api/resources_portal/management/commands/material_additional_metadata_key_change.py
```diff
@@ -0,0 +1,183 @@
...
+logger = get_and_configure_logger(__name__)
+
+
+def validate_category(category):
+    return len([choice for choice in Material.CATEGORY_CHOICES if category in choice]) > 0
+
+
+def paginate_queryset(query_set, function, pagination):
+    current_page = 0
+    size = len(query_set)
```
kurtwheeler, line 16:
```suggestion
    size = query_set.count()
```
This will use SQL's `COUNT` function so you don't have to load the whole queryset into memory (the point of pagination).

api/resources_portal/management/commands/material_additional_metadata_key_change.py
```diff
@@ -0,0 +1,183 @@
...
+    current_slice = []
+    materials_updated = 0
+
+    while current_page < size:
+        if (size - current_page) < pagination:
+            current_slice = query_set[current_page : current_page + (size - current_page)]
```
kurtwheeler, line 22:
```suggestion
            current_slice = query_set[current_page:]
```
`current_page : current_page + (size - current_page)` == `size`, but using a trailing `:` will get the rest of the list.

api/resources_portal/management/commands/material_additional_metadata_key_change.py
```diff
@@ -0,0 +1,183 @@
...
+            current_page += size - current_page
```
kurtwheeler, line 23:
```suggestion
            current_page = size
```
`current_page + size - current_page` == `size`

api/resources_portal/management/commands/material_additional_metadata_key_change.py
```diff
@@ -0,0 +1,183 @@
...
+
+logger = get_and_configure_logger(__name__)
+
+
+def validate_category(category):
+    return len([choice for choice in Material.CATEGORY_CHOICES if category in choice]) > 0
+
+
+def paginate_queryset(query_set, function, pagination):
+    current_page = 0
```
kurtwheeler, line 15: Could we change `current_page` to `current_index`? `current_page` makes me think it's whether we're on the first page or the second page, etc, with each one have `pagination` items in it.

kurtwheeler: Oh and maybe `pagination` -> `page_size`.

api/resources_portal/management/commands/material_additional_metadata_key_change.py
```diff
@@ -0,0 +1,183 @@
...
+        materials_updated += len(current_materials)
+        logger.info(
+            f"{len(current_materials)} materials processed. {materials_updated} total processed so far."
+        )
+
+    return materials_updated
+
+
+def log_nonapplicable_materials(query_set, key):
+    if len(query_set) > 0:
```
kurtwheeler, line 44:
```suggestion
    if query_set.count() > 0:
```

api/resources_portal/management/commands/material_additional_metadata_key_change.py
```diff
@@ -0,0 +1,183 @@
...
+        logger.warning(
+            f"The key '{key}' was not present in {len(query_set)} materials. The IDs of these materials are: "
```
kurtwheeler, line 46:
```suggestion
            f"The key '{key}' was not present in {query_set.count()} materials. The IDs of these materials are: "
```

api/resources_portal/management/commands/material_additional_metadata_key_change.py
```diff
@@ -0,0 +1,183 @@
...
+    if category:
+        query_set = Material.objects.filter(category=category)
+    else:
+        query_set = Material.objects.all()
+
+    filtered_query_set = query_set.filter(additional_metadata__has_key=old_key)
+    not_applicable_query_set = query_set.exclude(additional_metadata__has_key=old_key)
+
+    # Evaluates the queryset. Querysets are usually lazy, this forces evaluation
+    len(not_applicable_query_set)
```
kurtwheeler, line 90: Why do we need the queryset evaluated?

api/resources_portal/management/commands/material_additional_metadata_key_change.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def material_additional_metadata_key_change(category, old_key, new_key, pagination):
+    query_set = None
+    materials_changed = 0
+
+    if category:
+        query_set = Material.objects.filter(category=category)
+    else:
+        query_set = Material.objects.all()
```
kurtwheeler, line 110: This is repeated in each of these functions. Why not have this function just take the queryset as a parameter, so you can do this filtering once. If you wanted, you could even just have it take a pre-paginated queryset, although that'd be a bigger refactor.

api/resources_portal/management/commands/material_additional_metadata_key_change.py
```diff
@@ -0,0 +1,183 @@
...
+            current_page += size - current_page
+        else:
+            current_slice = query_set[current_page : current_page + pagination]
+            current_page += pagination
+
+        current_materials = []
+        for material in current_slice:
+            current_material = function(material)
+            current_material.updated_at = timezone.now()
+            current_materials.append(current_material)
```
kurtwheeler, line 32: I don't think you need to create a new list here. You can just pass the queryset to `bulk_update` and it doesn't look like you need to keep track of the return value of `function` because they just update the object that gets passed into them.

api/resources_portal/management/commands/material_additional_metadata_key_change.py
```diff
@@ -0,0 +1,183 @@
...
+    else:
+        query_set = Material.objects.all()
+
+    filtered_query_set = query_set.filter(additional_metadata__has_key=old_key)
+    not_applicable_query_set = query_set.exclude(additional_metadata__has_key=old_key)
+
+    # Evaluates the queryset. Querysets are usually lazy, this forces evaluation
+    len(not_applicable_query_set)
+
+    def add_func(material):
```
kurtwheeler, line 92: This didn't get changed for being in `remove_key`, and `material_additional_metadata_key_change` has an `add_func` as well.

api/resources_portal/management/commands/material_additional_metadata_key_change.py
```diff
@@ -0,0 +1,183 @@
...
+    def handle(self, *args, **options):
+        category = options["category"]
+        if category is not None:
+            if not validate_category(category):
+                raise CommandError(
+                    f"The provided category '{category}' is not a valid material category."
+                )
+
+        # Default query pagination is 100
+        pagination = 100
```
kurtwheeler, line 167: This is actually on the low side, I think a page of 1000 would be fine!

kurtwheeler (changes requested):
I've left a bunch more comments.

api/resources_portal/management/commands/material_additional_metadata_key_change.py
Philinphiladelphia: I tested this out, and I'm not sure if it's accurate. The `function` just modifies that material object currently loaded into memory, whereas the queryset is reflective of what's in the database, not in memory. I don't think it's especially computationally expensive to keep track of the data of a single slice of the materials in RAM at a time.

api/resources_portal/management/commands/material_additional_metadata_key_change.py
Philinphiladelphia: Good idea. I moved the category parsing into the main function body.

api/resources_portal/management/commands/material_additional_metadata_key_change.py
Philinphiladelphia: When using .count(), we don't. I refactored to remove this.

api/resources_portal/management/commands/material_additional_metadata_key_change.py
Philinphiladelphia: Done!

api/resources_portal/management/commands/material_additional_metadata_key_change.py
Philinphiladelphia: done!

api/resources_portal/management/commands/material_additional_metadata_key_change.py
Philinphiladelphia: Good ideas, much clearer. Done!

api/resources_portal/management/commands/material_additional_metadata_key_change.py
Philinphiladelphia: lol wow. I need to take some more courses in arithmetic. Done!

api/resources_portal/management/commands/material_additional_metadata_key_change.py
Philinphiladelphia: Sweet. Done!

api/resources_portal/management/commands/material_additional_metadata_key_change.py
Philinphiladelphia: Good catch. Changed to use this.

api/resources_portal/management/commands/material_additional_metadata_key_change.py
Philinphiladelphia: Good point. Fixed.

api/resources_portal/management/commands/material_additional_metadata_key_change.py
Philinphiladelphia: Changed to 1000!

kurtwheeler (approved):
Awesome! This looks great now!

api/resources_portal/management/commands/material_additional_metadata_key_change.py
kurtwheeler: Ok, works for me.

Issue #635, opened 2020-11-13
Request: Re-assign notifications go to the wrong user


dvenprasad:
### Context

`Not Deepa` re-assigned a request to `Chickpea Beans`.

`Not Deepa` got a notification (both email and in-app) that they were assigned to the request

<img width="750" alt="Screen Shot 2020-11-12 at 9 14 07 PM" src="https://user-images.githubusercontent.com/15315514/99020221-06ee4780-252c-11eb-8978-89d4771532f0.png">

<img width="618" alt="Screen Shot 2020-11-12 at 9 14 55 PM" src="https://user-images.githubusercontent.com/15315514/99020269-22f1e900-252c-11eb-8211-ebc4b3302f79.png">

### Problem or idea

These notifications should go to `Chickpea Beans` the assignee.

### Solution or next step

_You can tag others or simply leave it for further investigation, but you must propose a next step towards solving the issue._

PR #665, opened 2020-11-18
Refactors material_request update function so notifications have the right targets.

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/resources-portal/issues/635

We were sending notifications to the previous `assigned_to` instead of the updated `assigned_to`. While trying to fix it I realized that cleaning up the code and the order it does things would both fix the issue and make things easier to understand. Now we make a deep copy of the old object, update the object, and then compare the two. Nice and straight forward.

## Types of changes

## Functional tests

I added a test to check that the new user is actually who gets put on the notification.

## Checklist


davidsmejia (approved):
LGTM besides gh actions

PR #666, opened 2020-11-18
Fix additional weekly notification HTML issues

kurtwheeler:
## Issue Number

#644 

## Purpose/Implementation Notes

I either missed these messages or I was validating the HTML at the wrong step (maybe I had an empty list?). Now the validator only complains about attribute errors and the HTML page looks pretty good.

## Types of changes

- Bugfix (non-breaking change which fixes an issue)

## Functional tests

I printed out the inlined HTML, validated it, and have attached a screenshot of what it looks like in the browser.

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules

## Screenshots

image

PR #667, opened 2020-11-19
Now that weekly emails are working, go back to only sending them on Mondays

kurtwheeler:
## Issue Number

#666 

I want to check tomorrow's email before merging this, but I think we should be good. (Today's email was missing images, but that's because we weren't hosting them for some reason. This has been fixed.)


davidsmejia (approved):
🤖

PR #668, opened 2020-11-19
use updated api format for email invitations

davidsmejia:
## Issue Number

#630 

## Purpose/Implementation Notes

Fixes the call to the api which was previously `api.invite` to now match the crud format `api.invitations.create`

## Types of changes

- Bugfix (non-breaking change which fixes an issue)

## Functional tests

N/a

## Checklist

- [X] Lint and unit tests pass locally with my changes

## Screenshots

n/a

PR #669, opened 2020-11-19
Fix invitation emails

kurtwheeler:
## Issue Number

#668 

I got an invitation :tada:  It didn't have the images :sob: This should fix that up, it looks like it got change halfway, but some old code didn't get cleaned up and that caused the `REPLACE_` tags to get replaced with bad URLs.

## Types of changes


davidsmejia (approved):
🚀

PR #670, opened 2020-11-19
Fix HTML for invitation emails

kurtwheeler:
## Issue Number

#669 

## Purpose/Implementation Notes

Problem:
image

I think this should fix it. The validator only complains about CSS and attributes, and it looks solid in a browser. See below screenshot.

## Types of changes

- Bugfix (non-breaking change which fixes an issue)

## Functional tests

I took a screenshot.

## Checklist

- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules

## Screenshots

image
(missing images expected)

PR #671, opened 2020-11-19
More fixes for email invitation HTML

kurtwheeler:
## Issue Number

#670 

I missed a validation complaint that was actually relevant. I think this should fix it, but if not I'll try more tomorrow!

## Types of changes

## Functional tests

I've run email validation on the HTML that gets sent. It looked good in a browser before, but it still does too.


davidsmejia (approved):
🚀

PR #677, opened 2020-11-20
client two apps two environments

davidsmejia:
This PR removes the config files and github actions to manually deploy the application to production and staging instead:

* The client production app and staging app with be located on now.sh via resources-portal and resources-portal-staging respectively.
* The environment variables will be defined via docker.env for local and on hosted env's be located in the appropriate app's environment variables tabs on now.sh


kurtwheeler (approved):
This LGTM!

PR #580, opened 2020-10-27
misc email format changes

davidsmejia:
## Issue Number

#526 

Just testing some formatting changes

PR #679, opened 2020-11-30
Fix weekly notification email, increase frequency.

kurtwheeler:
## Issue Number

#580 

I think this should work now. It actually makes it through to the inlined email and modifying the source in my browser made the emails look right.

I upped the frequency to every hour. If it works I'll immediately bump it back down.


davidsmejia (approved):
LGTM

Issue #313, opened 2020-08-17
Deploy production


kurtwheeler:
### Context

We have a staging server, but not a production server.

### Problem or idea

We also need a production server, sometime around the end of this sprint.

### Solution or next step

Make sure the configuration works for prod and get the prod stack up and running.

PR #680, opened 2020-11-30
Add tfvars file for prod deploys

kurtwheeler:
## Issue Number

#313 

For now leave these as the same, we can scale up as necessary. This should make the deploy succeed.


davidsmejia (approved):
⛵

PR #681, opened 2020-11-30
Fix race condition in deploys

kurtwheeler:
## Issue Number

#679 

Looks like I added a race condition and some timings changed. I think I was relying on the API not coming up too fast, which must have changed.

## Types of changes


davidsmejia (approved):
LGTM

PR #684, opened 2020-12-01
Fix deploy issue with stopped containers 

kurtwheeler:
## Issue Number

#681 

Check for stopped containers too!


davidsmejia (approved):
LGTM!

Issue #686, opened 2020-12-01
Address Zip Code is required shouldnt be


davidsmejia:
### Context

Was just adding a schema validation for misc models that we post to the api and I think we should make zip codes optional since they arent universal.

PR #691, opened 2020-12-03
Makes postal_code optional.

kurtwheeler:
## Issue Number

#686 

Changes the field to null=True and adds a migration.


davidsmejia (approved):
Noice

PR #695, opened 2020-12-07
Make weekly emails weekly again

kurtwheeler:
## Issue Number

N/A, David just showed me a screenshot and it was easy

This was a mistake, but the fix is easy.

## Types of changes


davidsmejia (approved):
🎸

PR #698, opened 2020-12-08
log out branch env var

davidsmejia:
## Issue Number

n/a just want to check this in the deploy logs will remove after this is in master


kurtwheeler (approved):
Sure sure!

PR #700, opened 2020-12-08
remove log

davidsmejia:
## Issue Number

n/a

Ok this should end the saga of a thousand deploys. Looks like there was a relatively new config option to specify a production branch... the documentation does not express this adequately from the error message.


kurtwheeler (approved):
Sweet!!

PR #702, opened 2020-12-08
one more log to see why prod env vars arent set

davidsmejia:
## Issue Number

n/a

checking which env vercel thinks the master branch is


kurtwheeler (approved):
Let's do it!

PR #704, opened 2020-12-08
removing the most recent logs

davidsmejia:
## Issue Number

n/a 

remove the latest logs - the manual deploys are using the correct env vars and stage now so yay


kurtwheeler (approved):
:+1:

PR #705, opened 2020-12-09
Adds the OAUTH_CLIENT_ID env var everywhere it needs to be.

kurtwheeler:
## Issue Number

#313 

Login failed because it had the wrong client ID. Turns out we just never set that everywhere it needed to be. This does that. I've also added the secret in github actions.

## Types of changes

## Functional tests

I'll need to deploy both staging and prod to test this.

## Checklist


davidsmejia (approved):
LGTM!

Issue #690, opened 2020-12-03
Account creation seems to be broken if ORCID email is private


kurtwheeler:
### Context

I tried creating an account locally with an ORCID account that had email settings set to private.

### Problem or idea

I got an error: `{"error": "invalid_grant", "error_description": "Reused authorization code: 19cgTF"}`

### Solution or next step

It looks like ORCID isn't allowing us to reuse it's authorization code we get from it, so if they don't have an email for us then we have to go through the flow again passing along the email the user provides us.

PR #709, opened 2020-12-17
Allow users to create account and sign in with limitied orcid public visibility

davidsmejia:
## Issue Number

#690 

* Combines the logic for signing and create account to use the same hook
* handles missing details depending on what comes back from orcid
* makes first name and last name required to create an account on the API
* refactors some of the components in favor of composition

## Types of changes

## Functional tests

Tested Sign In and Create Account flows with:

All Open Settings
Private Email
Private Name

## Checklist

## Screenshots

n/a


api/resources_portal/views/user.py
```diff
@@ -139,102 +167,81 @@ def get_permissions(self):
...
-        last_name = ""
-        if summary["person"]["name"].get("family-name", None):
-            last_name = summary["person"]["name"]["family-name"]["value"]
+        if missing_data:
+            return JsonResponse(
+                {
+                    "error": "There were details not provided on the provided ORCID record. Please provide missing details in the POST request.",
+                    "required": missing_data,
+                },
+                status=401,
```
kurtwheeler, line 222: I think a 400 Bad Request makes more sense here than a 401 Unauthorized.

client/src/components/LoginRequired.js
```diff
@@ -29,12 +31,22 @@ export default ({
...
+            ) : (
+              <CreateOrLogin />
+            )}
           </Box>
         </Main>
       </Box>
     )
   }
 
+  // as you were
```
kurtwheeler, line 50: What's this mean?

kurtwheeler (approved):
This looks good to me!! Nice work with the refactor of create user, that looks wayyyy better now!

api/resources_portal/views/user.py
davidsmejia: updated to use 409 when user exists and 422 when orcid response is missing required info

client/src/components/LoginRequired.js
davidsmejia: updated comment for clarity

kurtwheeler (approved):
The additional changes LGTM! Nice work!

PR #715, opened 2021-01-08
Check the additional metadata fields of imported resources. 

kurtwheeler:
## Issue Number

N/A came up while David was doing some testing with dev data.

Also add missing accession code. Also call the check when loading up dev data to make sure it's valid.

## Types of changes

## Functional tests

After changing the `validate_import` function, `rportal populate-db` was failing for me. Then I added the missing accession code and it worked happily again.

## Checklist


davidsmejia (approved):
🚀  nice!

Issue #616, opened 2020-11-06
Manage Resources: Cannot view closed requests


dvenprasad:
### Context

On the manage resource page for a specific resource, clicking on the view closed requests does nothing. 

### Problem or idea

It show users closed requests. 

### Solution or next step

_You can tag others or simply leave it for further investigation, but you must propose a next step towards solving the issue._

PR #718, opened 2021-01-11
616 cannot view closed requests

davidsmejia:
## Issue Number

#616 

uses the correct option names for modals also adds some spacing

## Types of changes

## Functional tests

Tested locally

## Checklist

## Screenshots

Screen Shot 2021-01-11 at 6 06 46 PM


kurtwheeler (approved):
This LGTM!

I will say that it might make sense for the open requests to be listed about the closed ones though.

davidsmejia:
> I will say that it might make sense for the open requests to be listed about the closed ones though.

It's sorted recent to latest which I think is ok for now, we should probably add a sort by status, thought this would have to be on the frontend since there are more statuses on the FE than the BE

Issue #645, opened 2020-11-13
When requests go to an individual, use person's name


dvenprasad:
### Context

When the resources are owned by a individual rather than a team, messaging across the platform refers to them as `My Resources` 
<img width="569" alt="Screen Shot 2020-11-13 at 2 41 03 PM" src="https://user-images.githubusercontent.com/15315514/99122401-3f456280-25cc-11eb-9ee0-a407c8a43018.png">

<img width="762" alt="Screen Shot 2020-11-13 at 2 48 28 PM" src="https://user-images.githubusercontent.com/15315514/99122435-53895f80-25cc-11eb-93ec-90f8730e7421.png">

<img width="680" alt="Screen Shot 2020-11-13 at 2 37 39 PM" src="https://user-images.githubusercontent.com/15315514/99122471-613ee500-25cc-11eb-929b-59c186f2c821.png">

### Problem or idea

We should the person's name instead. 

### Solution or next step

_You can tag others or simply leave it for further investigation, but you must propose a next step towards solving the issue._

PR #723, opened 2021-01-12
645 naming personal org My Resources was a bad idea

davidsmejia:
## Issue Number

#645 

So it was probably a bad idea to rename the personal org's `My Resources`. It seems more prudent to rewrite the form drop down when listing/importing a resource (one location) than modifying it whenever we send out a notification or a when the requester fields the requirements/contacts/waits for action from the org (this happens in like a million places).

 - sets the default name of the personal org to the user.full_name
 - This adds a helper for taking a single org or list of orgs and renames anything where is_personal_organization is true to 'My Resources' (this only happens when listing the resources.
 - implements the helper in the list/add drop down organization picker component
 - removes unneeded code which the above changes handle

We could probably add a migration for this change since the staging db will still look wonky. But all new personal organizations will be named after their owner. We could also one time change the db on staging.

## Types of changes

## Functional tests

tested listing and managing a resource request from both sides of the process

## Checklist

## Screenshots

n/a

### DIFF
--- a/api/resources_portal/test/views/test_user.py
+++ b/api/resources_portal/test/views/test_user.py
@@ -71,7 +71,7 @@ class TestUserPostTestCase(APITestCase):
 
         self.assertEqual(response.status_code, status.HTTP_200_OK)
 
-        self.assertEqual(user.personal_organization.name, "My Resources")
+        self.assertEqual(user.personal_organization.name, user.full_name)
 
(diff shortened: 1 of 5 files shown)


kurtwheeler (approved):
Sweet, this does indeed look like a better way to handle this.

If you want, I think I could just exec onto the staging API and run a bit of django code to update all the personal orgs to have their owners names.

It's not ideal, but it's just staging and won't be dangerous.

Issue #652, opened 2020-11-16
Teams: Let users leave team


dvenprasad:
### Context

Reported by Anna.

### Problem or idea

Users should be able to leave team like so 👇 

<img width="754" alt="Screen Shot 2020-11-16 at 3 37 53 PM" src="https://user-images.githubusercontent.com/15315514/99305401-b889c300-2821-11eb-8319-ad5bfc679982.png">

### Solution or next step

_You can tag others or simply leave it for further investigation, but you must propose a next step towards solving the issue._

PR #725, opened 2021-01-12
652 let team members leave team

davidsmejia:
## Issue Number

#652 

We will probably need to add some logic to the api that checks if the user is a point person for any requests and automatically reassigns those to the owner.

This implements the button on the side of the team members table for letting people leave a team.

## Types of changes

## Functional tests

added myself to a team and left the team

## Checklist

## Screenshots

Screen Shot 2021-01-12 at 11 18 41 AM

Screen Shot 2021-01-12 at 11 19 16 AM

### DIFF
--- a/client/src/components/TeamMembers.js
+++ b/client/src/components/TeamMembers.js
@@ -9,7 +9,15 @@ import { useAlertsQueue } from 'hooks/useAlertsQueue'
 
 export default () => {
   const { addAlert } = useAlertsQueue()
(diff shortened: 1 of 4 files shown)


kurtwheeler (approved):
This LGTM!!!

Regarding 
> We will probably need to add some logic to the api that checks if the user is a point person for any requests and automatically reassigns those to the owner.

I'm way ahead of you: https://github.com/AlexsLemonade/resources-portal/blob/dev/api/resources_portal/views/organization_member.py#L50 ;)

Issue #634, opened 2020-11-13
Edit Resource: Show specified request requirements


dvenprasad:
### Context

When user wants to edit a resource's request requirement, they are presented with this:
<img width="780" alt="Screen Shot 2020-11-12 at 9 04 23 PM" src="https://user-images.githubusercontent.com/15315514/99019609-aa3e5d00-252a-11eb-8e54-2984c0cad113.png">

### Problem or idea

- There's no indication which requirements are currently applied. 

- Does not have affordance to edit those requirements, the UI implies that users can only specify new requirements. 

- When users click on `Specify new requirements` they see the below screen, which is not the promise that button makes.
<img width="775" alt="Screen Shot 2020-11-12 at 9 03 22 PM" src="https://user-images.githubusercontent.com/15315514/99019865-53855300-252b-11eb-976b-44d260158f4d.png">

Recommendation: 
We should show what we show under `Specify New Requirements` and have the heading be `Edit Requirements` 

### Solution or next step

_You can tag others or simply leave it for further investigation, but you must propose a next step towards solving the issue._

PR #727, opened 2021-01-12
634 update edit requirements

davidsmejia:
## Issue Number

#634 

So the main thing this PR does is correctly anticipate the version of the form to show when it is presented to a user:

For New resources show them the picker if they have something to pick from otherwise the regular form
For Editing Show Them the form by default unless it is unsaved and they picked a resource to copy requirements from

These are the form/picker titles
 - `Edit Request Requirements` (form when saved)
 - `Specify New Requirements` (form when unsaved)
 - `Copy Requirements from Another Resource` (picker button)

The only outlier is the title is not the same as the button for the picker
`Use Same Requirements As` ( picker title)

Also updated the `getRequestRequirements(resource)` helper to handle unsaved resources
Also added an `isSaved` export from useResourceForm to help with determining mode.

@dvenprasad I changed the copy from `Select Existing Resource` because the meaning is ambiguous when you are editing a saved (existing) resource.

## Types of changes

## Functional tests

Tested new, edited, edit unsaved modes with new and existing requirements

## Checklist

## Screenshots

Screen Shot 2021-01-12 at 5 12 06 PM

### DIFF
--- a/client/src/components/resources/RequestRequirementsForm.js
+++ b/client/src/components/resources/RequestRequirementsForm.js
@@ -263,15 +263,15 @@ export default () => {
             options={teamResourceOptions}
             name="existing-requirements"
           >
-            {({ resource, ...option }, { hover }) => (
+            {({ resource: existingResource, ...option }, { hover }) => (
               <RadioButton
                 key={option.label}
                 name={option.name}
                 hover={hover}
                 checked={existingRequirementsResource.id === option.value}
                 value={option.value}
-                label={<RequirementsLabel resource={resource} />}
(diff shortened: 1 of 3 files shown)


dvenprasad:
Okay, yeah these CTAs sound good to me. 

>The only outlier is the title is not the same as the button for the picker
Use Same Requirements As ( picker title)

I am okay with changing `Use Same Requirements As` ----> `Copy Requirements from Another Resource` to not confuse users.

kurtwheeler (approved):
LGTM!

Issue #619, opened 2020-11-06
Search Cards: Labels missing


dvenprasad:
### Context

On request cards the following fields are labels:
- Publication / Pre-print
- Request Requirement

### Problem or idea

Add them. 

Also, these fields show be shown even if they are empty. If they are empty, then we should show None. 

### Solution or next step

_You can tag others or simply leave it for further investigation, but you must propose a next step towards solving the issue._

davidsmejia:
@dvenprasad Would you be able to send me a screen shot please? I'm not 100% sure which request card you're referring to.

dvenprasad:
Oh.... ohhh.... oh!  Sorry my bad.... I meant search cards. 


![Screen Shot 2021-01-13 at 9 26 53 AM](https://user-images.githubusercontent.com/15315514/104464945-7e1e7700-5581-11eb-8c9b-570d301598d8.png)

dvenprasad:
Also, the `Link to Publication` should be whatever value user provides for `Publication Title` or `Pre-print Title` 

![Screen Shot 2021-01-13 at 1 59 37 PM](https://user-images.githubusercontent.com/15315514/104496998-b89b0a80-55a7-11eb-933c-6150f83c534d.png)


![Screen Shot 2021-01-13 at 2 00 02 PM](https://user-images.githubusercontent.com/15315514/104497017-c05aaf00-55a7-11eb-95d0-a21f5aa85376.png)

For Imported Resources:

![Screen Shot 2021-01-13 at 2 00 25 PM](https://user-images.githubusercontent.com/15315514/104497059-cfd9f800-55a7-11eb-9c6e-4002594c3a18.png)

davidsmejia:
closed with #736

PR #730, opened 2021-01-13
619 - missing lables and publication/pre-print titles on search cards

davidsmejia:
## Issue Number

#619 

Adds back the labels and looks for the publication title and pre-print in the on the material directly (no longer nested in additional metadata)

I think that the dev data way back when wasn't updated and that was causing this to be an issue.

## Types of changes

## Functional tests

n/a locally works

## Checklist

## Screenshots
Bad screen shot because I never fill out optional data but we can see it on staging.

Screen Shot 2021-01-13 at 3 52 17 PM

### DIFF
--- a/client/src/components/resources/SearchResult.js
+++ b/client/src/components/resources/SearchResult.js
@@ -97,7 +97,8 @@ export const SearchResultDetail = ({
   const joinedValue = Array.isArray(data.value)
     ? data.value.join(', ')
     : data.value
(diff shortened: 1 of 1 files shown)


dvenprasad:
But will the publication title show up instead of `Link to Publication`?

davidsmejia:
> But will the publication title show up instead of `Link to Publication`?

yeah it will only default to that if there isn't a title to show for the link

dvenprasad (approved):
🐢

Issue #615, opened 2020-11-06
Manage Resources: More Options -> View Resource redirects to 404


dvenprasad:
### Context

On the resource cards under Manage Resources, View resource redirects to 404.

### Problem or idea

It should take me to the `View Resource` page.

### Solution or next step

_You can tag others or simply leave it for further investigation, but you must propose a next step towards solving the issue._

PR #736, opened 2021-01-14
615 - fix view resource url in manage resource card

davidsmejia:
## Issue Number

#615

Notes:
*Makes the path go to the resource instead of the nested url


kurtwheeler (approved):
Subtle :+1:

Issue #298, opened 2020-08-05
Material Request Additional Documents


davidsmejia:
We need to prevent the requester from uploading the documents piece meal after their material_request has been accepted 

It would be great to validate the update on material request when the requester tries to attach their IRB, partially executed MTA, and shipping codes - we ensure that they are all sent in the same request.

kurtwheeler:
@davidsmejia is this still relevant or did we end up addressing this in some manner?

davidsmejia:
> @davidsmejia is this still relevant or did we end up addressing this in some manner?

Looks like this was taken care of! https://github.com/AlexsLemonade/resources-portal/blob/dev/api/resources_portal/views/material_request.py#L104

Cant find the PR but here is the commit
https://github.com/AlexsLemonade/resources-portal/commit/46d6604cd84cab2289a3a84d10f3fc784ae2cfea

Closing since this was addressed.

davidsmejia:
Actually this is still relevant since we do not check that any of the updates are completely updated.
For example if you need the IRB and the partially executed MTA this issue is saying we should force the user to post them at the same time instead of allowing one then the other. This helps simplify a lot of the logic for determining the current state of the request.

I'm not 100% sure at the moment if doing that would break the client - seems possible. But I do know the client does not allow the user to do that. So this is mainly a nice to have for overall system integrity. Meaning the API backs up what the client is coercing the user to do.

PR #740, opened 2021-01-20
Force MTA and IRB to be uploaded at the same time if both are required.

kurtwheeler:
## Issue Number

#298 

If one has been uploaded without the other, it's hard to determine what state the request is in. This forces them to be uploaded together.

## Types of changes

## Functional tests

I added unit tests to cover this.

## Checklist

### DIFF
--- a/api/resources_portal/views/material_request.py
+++ b/api/resources_portal/views/material_request.py
@@ -494,17 +494,31 @@ class MaterialRequestViewSet(viewsets.ModelViewSet):
             mta_changed = field_changed(
                 "requester_signed_mta_attachment", original_material_request, material_request
             )
+            payment_changed = field_changed(
+                "payment_method", original_material_request, material_request
+            )
+            payment_notes_changed = field_changed(
+                "payment_notes_method", original_material_request, material_request
+            )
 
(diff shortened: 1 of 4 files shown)


api/resources_portal/views/material_request.py
```diff
@@ -488,17 +488,53 @@ def update(self, request, *args, **kwargs):
...
+                    missing_document = True
+                if original_material_request.material.needs_mta and not mta_changed:
+                    missing_document = True
+                if (
+                    original_material_request.material.shipping_requirement
+                    and original_material_request.material.shipping_requirement.needs_payment
+                ):
+                    if not (payment_changed or original_material_request.payment_method) or (
+                        payment_notes_changed or original_material_request.payment_notes_method
+                    ):
```
davidsmejia, line 520: I think that the original_material_request should always have these be empty at least it should be if we are enforcing these changes.

kurtwheeler: I agree, but we haven't until now and this just seems overall safer than assuming they are.

davidsmejia (approved):
LGTM!

Issue #708, opened 2020-12-16
PDX list form: Cannot add resource


dvenprasad:
### Context

Tried to add a PDX. It said that following fields were not filled even though they were:

- Passage QA Performed
- Is strain immunized
- Is from untreated patient

### Problem or idea

Also the following fields need to marked optional:

- Treatment Drug
- Treatment Protocol
- Response to Prior Treatment
- Virology Status
- Tumor Sample Type
- Existing Model Explaination
- Treatment for Engraftment
- Engraftment Rate
- Engraftment Rate
- Response to Standard of Care
- Animal Health Status
- All fields in `Model Study` section
- Number of Available Models
- PDX Model Availability 
- Governance Restriction

PR #742, opened 2021-01-21
708 pdx list form fix configuration

davidsmejia:
## Issue Number

#708

Well this kinda got out of hand.

This PR:
* lets PDX's get listed ( by fixing the validation and correctly setting the attribute datatypes)
* improves ui perf by debouncing updates to the underlying resource/material object
* fixes presentation and validation of boolean values
* moves reused logic to helpers
* allows optional attributes to be inferred from the validation schema for a partially defined material/resource
* small refactor for presenting form fields in a more stream line fashion
* moves smaller one time components to their own files for clarity
* fixes issue where resetting the resource on the "Success" page was throwing an error

## Types of changes

## Functional tests

Listed and previewed PDXs locally

## Checklist

## Screenshots

n/a


client/src/helpers/getOptionalAttributes.js
```diff
@@ -0,0 +1,18 @@
+// converts a schema to a list of optional attributes
+export default (schema, object) => {
+  const optionalAttributes = []
+
+  const addRequired = ([key, desc]) => {
```
kurtwheeler, line 5: Is there a reason this takes a list and deconstructs it instead of just taking the args directly?

client/src/schemas/material.js
```diff
@@ -67,56 +78,8 @@ const importedCategories = {
...
-    tumor_omics: string(),
-    metastases_in_strain: string(),
-    lag_time: string(),
-    number_of_available_models: string(),
-    pdx_model_availability: string(),
-    governance_restriction: string()
-  }),
+  // PDXs currently can't be imported
+  PDX: object(),
   OTHER: object({
```
kurtwheeler, line 83: Is this intentional? It looks like PDX is an empty object and OTHER is defined twice.

davidsmejia: 1) Yeah it just means additional attributes should always be an object but doesnt validate for keys. This is in case import is toggled. It wont validate but I just kept the prop there. 
2) it just looks like that from the code folding. The `OTHER:` here is for imported materials the next one is for listed.

client/src/helpers/getOptionalAttributes.js
davidsmejia: Yeah this is more of a design choice since its defined inside the helper and used recursively I can do this which passes the pair in.
I prefer this:
```javascript
    if (desc.fields) Object.entries(desc.fields).forEach(addRequired)
```
Over this:
```javascript
    if (desc.fields) Object.entries(desc.fields).forEach(([k, v]) => addRequired(k, v))
```
```

client/src/schemas/material.js
kurtwheeler: Oops that was just cold folding. Good call :+1:

kurtwheeler (approved):
Ok sweet, this LGTM!

PR #746, opened 2021-02-01
break up manage request forms ( requester form )

davidsmejia:
## Issue Number

N/A

Breaks up the requester manage request form into smaller more manageable (lol) pieces.

## Types of changes

## Functional tests

tested the requester side of things. should work the same way

## Checklist

## Screenshots

n/a


client/src/components/RequestHero.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react'
```
kurtwheeler, line 1: RequestHero? What does that mean?

kurtwheeler (approved):
This overall LGTM, although I had one question about a name. Looks way cleaner and smaller files, I love it!

client/src/components/RequestHero.js
davidsmejia: A hero is like a page header that gives you context before the fold usually with one or two CTA's. I do think this is a bit of a stretch here - I was trying to avoid something like `RequestResource` though maybe something like `RequestResourceDetails` might be better.

davidsmejia: ok upated to `RequestResourceContact`

PR #754, opened 2021-02-08
break up manage request forms (sharer form)

davidsmejia:
## Issue Number

N/A

* Adds more smaller components for better composability.
* Combines the two request forms into one smaller file so both forms have identical state management and less code duplication.
* Adds more functionality to the `useRequest` hook
* overall easier to see whats going on in that page and easier to add/remove functionality
* fixes some bugs discovered along the way

## Types of changes

## Functional tests

n/a tested with lots of different request requirements

## Checklist

## Screenshots

n/a


kurtwheeler (approved):
This LGTM! This seems like a much more managable way to deal with all the states that a request could be in. Nice work!

PR #759, opened 2021-02-09
prepare for hubspot invites

davidsmejia:
## Issue Number

n/a

* Adds `grant.year`
* Adds a filter when parsing the grant data from the url to remove any objects with empty values ( this happens when hubspot sends emails)

## Types of changes

## Functional tests

n/a

## Checklist

## Screenshots

n/a


kurtwheeler (changes requested):
I think you need to add this to the grant serializers [here](https://github.com/AlexsLemonade/resources-portal/blob/dev/api/resources_portal/serializers/relation_serializers.py#L108) and [here](https://github.com/AlexsLemonade/resources-portal/blob/dev/api/resources_portal/views/grant.py#L11)

davidsmejia:
> I think you need to add this to the grant serializers [here](https://github.com/AlexsLemonade/resources-portal/blob/dev/api/resources_portal/serializers/relation_serializers.py#L108) and [here](https://github.com/AlexsLemonade/resources-portal/blob/dev/api/resources_portal/views/grant.py#L11)

good catch!

kurtwheeler (approved):
LGTM!

Issue #595, opened 2020-10-29
Notification emails button is aligned to the left and I don't know why


kurtwheeler:
### Context

#526 is mostly fixed. The emails now render the same in all email clients.

### Problem or idea

![image](https://user-images.githubusercontent.com/4026833/97582112-c3dc9200-19cb-11eb-93a6-d7d8fb4a3027.png)

The call to action button is shifted to the left.

### Solution or next step

I've looked into this as far as I really know how. When the inliner is run, it adds a lot of `align='left'` to elements that don't have it in the source. I assume it must be from a CSS style or something, but I cannot deduce what is doing it. My guess is this will be pretty trivial for someone who has worked in depth with CSS more recently.

PR #772, opened 2021-02-12
595 - center notification email CTA

davidsmejia:
## Issue Number

#595 

* wraps the CTA in a full with table and centers the contents
* removes the brs since they werent being applied
* indents the main text variable because it looks nicer but wont change how the html is rendered

## Types of changes

## Functional tests

Will need to test on staging once this is approved.

## Checklist

## Screenshots

n/a


kurtwheeler (approved):
OoooooOOoooohhhhh. Nice find! I'm so glad this is finally sorted :joy:

Issue #515, opened 2020-10-09
Sentry dialog box opens once per error instead of once per page


kurtwheeler:
### Context

We've got a known issue with notifications

### Problem or idea

When I go to list my notifications I get the error dialog box with a black border. When I click close, it stays, except the border gets less dark each time I click it until it's gone.

### Solution or next step

I think it's because it opens up one dialog box per error, which is probably one per notification. Ideally we'd only open one box per page and/or interaction.

kurtwheeler:
@davidsmejia this is done correct?

davidsmejia:
> @davidsmejia this is done correct?

No this is not done. I think we will want to roll out our own version of the SentryErrorBoundary and handle that. I am not sure at the moment how we would be able to do that and keep the dialogue box though.

PR #773, opened 2021-02-12
515 sentry has too many dialogue boxes

davidsmejia:
## Issue Number

#515 

Sentry kept opening new dialogue boxes most likely because it did not have the fallback set up correctly.
I think since the fallback page wasnt being rendered the errors kept accumulating in the error boundary.
This PR corrects that implementation and cleans up the fallback page/component.
Also when an error is encountered it navigates to to the previous page (presumably one without an error) and hard refreshes the page. We will need to test this in stage to see if it suits out needs but it seems to be good enough.

Tagging @dvenprasad for copy check.

## Types of changes

## Functional tests

tested with deliberate errors

## Checklist

## Screenshots

Screen Shot 2021-02-12 at 4 30 41 PM

### DIFF
--- a/client/src/pages/_app.js
+++ b/client/src/pages/_app.js
@@ -16,6 +16,10 @@ const LoginRequired = dynamic(() => import('components/LoginRequired'), {
   ssr: false
 })
 
+const Fallback = (sentry) => {
+  return <Error sentry={sentry} />
+}
+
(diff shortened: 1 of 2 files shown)


kurtwheeler (approved):
Ok cool! This looks like it'll be way better!

Issue #614, opened 2020-11-06
Manage Resource: Anchor Open Request on card to open request section


dvenprasad:
### Context

![2020-11-06 17 29 36](https://user-images.githubusercontent.com/15315514/98420690-b66f7980-2055-11eb-8c49-a33cbf5332b7.gif)

### Problem or idea

When user click on `X Open Request` take them to `Open Requests` section on Manage Resource page. 

### Solution or next step

_You can tag others or simply leave it for further investigation, but you must propose a next step towards solving the issue._

PR #787, opened 2021-03-04
614 open requests to open requests section

davidsmejia:
## Issue Number

#614 

If a page has a component `<ScrollTo name="open-requests" />` then linking to the page with that component with the additional query param "?scroll=open-requests" will attempt to scroll that content into view.

This works best when the `<ScrollTo />` component is inside a div or Box with it's height at min 100vh so the content can scroll entirely to to the top. I think the way we navigate to content like this should probably be revisited since it is inconsistent with how the rest of the site works. Also the scrolling will be applied when the link is shared or backwards navigation occurs which adds complexity for determining when /if that is appropriate.

This solution works 99% of the time depending on how the api responds and if the contents can render within the first and would require a much more significant lift to change the parent passes data to child paradigm that has been adhered to.

Future thoughts:
We might be able to implement something like a `useScroll` that takes an element ref and returns a function to call when it is ready to be scrolled to.

## Types of changes

## Functional tests

n/a scrolls like 99% of the time based on a conservative delay

## Checklist

### DIFF
--- a/client/src/components/ScrollTo.js
+++ b/client/src/components/ScrollTo.js
@@ -18,8 +18,9 @@ export default ({ name, children }) => {
     if (nameMatch && hasRef && !scrolled) {
       scrolledRef.current = true
       // TODO: this will need to be updated when responsive to account for the
-      // also this is not the best implementation of this feature since the
-      // document height is not set until after the api requests are returned
+      // fixed height of the header. Also this is not the best implementation of
+      // this feature since the document height is not set until after the api
+      // requests are returned
       setTimeout(() => {
         window.scroll(0, ref.current.offsetTop - 83)
       }, 475)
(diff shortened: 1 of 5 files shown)


client/src/components/ScrollTo.js
```diff
@@ -0,0 +1,29 @@
...
+  const router = useRouter()
+  const ref = React.useRef()
+  const scrolledRef = React.useRef(false)
+  React.useEffect(() => {
+    const nameMatch = name && name === router.query.scroll
+    const hasRef = !!ref.current
+    const scrolled = scrolledRef.current
+    if (nameMatch && hasRef && !scrolled) {
+      scrolledRef.current = true
+      // TODO: this will need to be updated when responsive to account for the
```
kurtwheeler, line 20: for the what?

kurtwheeler (approved):
LGTM!

Issue #737, opened 2021-01-14
Add a mechanism for users without grant id to let us know they want to add resources


dvenprasad:
### Context
Currently, the portal does not allow users without a grant id to add resources or create teams. 

### Problem or idea

Add a button somewhere (My Account seems to make the most sense/ or on Add resources page) for them to let us know that they would like to be able to add resources. This would be good to collect information to understand when to prioritize #215 . 

### Solution or next step

Design and implement a mechanism for this after beta but before launch

davidsmejia:
Since users don't know that they can't add resources without one. I think we could also add a separate guard that prevents people from going to the list/import form pages, right now, they wont be able to get past the select grant step. If we catch them earlier in the process (we can check that they don't belong to an org, or if they do their orgs have no grants) we could present a button on the `/resources` page to collect that info.

dvenprasad:
The best places to provide them this button is when we require Grant ID to do a thing. There are three places
- Listing a resource
- Importing a resource
- Creating a team 

We can prompt at these places to let us know if they wish to list resources without a grant ID.  @davidsmejia  mentioned yesterday, we can do a fullscreen takeover and ask them to let us know. 

Maybe we should track this in a google/hubspot form instead of an email to grants team (harder to keep track). It maybe easier for us measure interest and to reach back out when we do let people list without grant ids. Am I over engineering this? @jaclyn-taroni

jaclyn-taroni:
> Maybe we should track this in a google/hubspot form instead of an email to grants team (harder to keep track). It maybe easier for us measure interest and to reach back out when we do let people list without grant ids. Am I over engineering this?

Would you couch it as sign up to be notified if we add this feature? It makes sense to me that I would enter my email if that was the prompt -> a list on HubSpot. I imagine if someone is highly motivated to draft an email that provides a lot of additional context that we may find useful, they would probably find a way to send that email if that makes sense.

dvenprasad:
Yeah I think getting their email would be one of the questions on the form. We can always provide a box for them to tell us things or ask for specific information we may want to know, like what do they want to share? (although I'm not sure how to phrase that last question with less scoopy vibes)

I was thinking of couching it as, hey we're measuring interest in providing this feature.

dvenprasad:
![list-resource-no-grant-id](https://user-images.githubusercontent.com/15315514/109862999-1acece00-7c2f-11eb-833d-9b8e93ac7e97.png)



[...]

PR #796, opened 2021-03-19
737 adding resource without grant

davidsmejia:
## Issue Number

#737 

* adds a generic hook to talking to hubspot api
* adds a `GrantsRequired` guard that forced the feature request form if user's account does not have any grants
* `GrantsRequired` runs after `AccountRequired` so as it stands right now they would have to login to find out they cant post without a grant.
* validates the input which is handled by the hook using generic helpers (very cool) and takes an arbitrary form schema
* alerts user to success/errors and redirects them home when they are done

NOTE: if the hubspot form updates, we will need to update the form component to match expected inputs... there was no way to use the embed code in a way that would suit our needs.

## Types of changes

## Functional tests

Tested validation, successful response, and bad response all seem to be working as expected

## Checklist

## Screenshots

Screen Shot 2021-03-19 at 4 05 35 PM
Screen Shot 2021-03-19 at 4 06 16 PM

Updated:

Screen Shot 2021-03-19 at 4 41 15 PM

### DIFF
--- a/client/src/components/GrantRequired.js
+++ b/client/src/components/GrantRequired.js
@@ -24,10 +24,7 @@ export default ({ children }) => {
       <>
         {children}
         <Modal showing nondismissable>
-          <ResourceAddNoGrantRequestForm
-            portalId="5187852"
-            formId="960eaa6e-dc9a-4921-874f-40ba9bae8583"
-          />
+          <ResourceAddNoGrantRequestForm />
         </Modal>
       </>
     )
--- a/client/src/components/GrantRequired.js
+++ b/client/src/components/GrantRequired.js
@@ -1,0 +1,37 @@
+import React from 'react'
+import ResourceAddNoGrantRequestForm from 'components/ResourceAddNoGrantRequestForm'
+import { Loader } from 'components/Loader'
+import { useUser } from 'hooks/useUser'
+
+import { Modal } from 'components/Modal'
+
+export default ({ children }) => {
+  const { isAbleToAddResources } = useUser()
+  const [userOk, setUserOk] = React.useState()
+
+  React.useEffect(() => {
+    const asyncCheck = async () => {
+      setUserOk(await isAbleToAddResources())
+    }
+    if (userOk === undefined) asyncCheck()
+  })
+
+  if (userOk === undefined) return <Loader />
+
+  // show feature request form if user does not have a grant
+  if (!userOk) {
+    return (
+      <>
+        {children}
+        <Modal showing nondismissable>
+          <ResourceAddNoGrantRequestForm
+            portalId="5187852"
+            formId="960eaa6e-dc9a-4921-874f-40ba9bae8583"
+          />
+        </Modal>
+      </>
+    )
+  }
+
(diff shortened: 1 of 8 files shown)


dvenprasad (approved):
Could we add. (optional) next to `Is there anything you would like to add` ?

Otherwise it looks good! 🐈 🐈

client/src/components/GrantRequired.js
```diff
@@ -0,0 +1,37 @@
...
+
+  // show feature request form if user does not have a grant
+  if (!userOk) {
+    return (
+      <>
+        {children}
+        <Modal showing nondismissable>
+          <ResourceAddNoGrantRequestForm
+            portalId="5187852"
+            formId="960eaa6e-dc9a-4921-874f-40ba9bae8583"
```
kurtwheeler, line 29: Are these magic strings from Hubspot?

kurtwheeler:
Hmm, is it possible that a lab member could be confused by this if they are trying to list or import a resource for their PI who does have an Alex's grant? The form doesn't mention that they may just need an invite to their PI's team.

client/src/components/GrantRequired.js
kurtwheeler: Oh I see it now. Maybe leave a comment or something? Hubspot isn't mentioned in this file anywhere.

client/src/hooks/useUser.js
```diff
@@ -110,6 +110,26 @@ export const useUser = (defaultUser, defaultToken) => {
...
+    )
+
+    if (!userIsOk) return false
+
+    // user has grants
+    const { grants, organizations: teams } = freshUser
+
+    const hasGrants = grants.length > 0
+    // join all the arrays of org grants to one array and check if empty
+    const hasTeamGrants = [].concat(...teams.map((t) => t.g)).length > 0
```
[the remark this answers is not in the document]
davidsmejia: ooph good catch

client/src/hooks/useHubspotForm.js
```diff
@@ -0,0 +1,67 @@
...
+    } catch (e) {
+      setErrors(getErrorMessages(e))
+      return false
+    }
+
+    return true
+  }
+
+  // can't just post the json data, no that would be too easy
+  // convert to {fields: [{name, value}]}
```
kurtwheeler, line 44: Hahahaha that's horrible!

client/src/components/GrantRequired.js
davidsmejia: ok i moved them to env vars

davidsmejia:
> Hmm, is it possible that a lab member could be confused by this if they are trying to list or import a resource for their PI who does have an Alex's grant? The form doesn't mention that they may just need an invite to their PI's team.

@dvenprasad might be able to give a response to that... they would only see this if they do not have a grant given to them when signing up or if they are not yet a member of a team with a grant

kurtwheeler (approved):
LGTM but we might want to add a bit of help for someone who's just trying to list a resource for their PI.

Issue #799, opened 2021-03-22
PATCHing material requests 400s


kurtwheeler:
### Context

From the cloudwatch logs:
```
[pid: 26|app: 0|req: 230/2214] 172.17.0.1 () {52 vars in 970 bytes} [Mon Mar 22 15:45:33 2021] PATCH /v1/material-requests/5/ => generated 66 bytes in 62 msecs (HTTP/1.1 400) 6 headers in 210 bytes (1 switches on core 0)
```

### Problem or idea

I'm not actually sure why this is 400ing... There's no additional info in the logs.

### Solution or next step

This needs to be investigated. Not sure what's going on yet. I was watching the usability eval and it was happening while he was trying to make a request. The logs have a successful POST to `/attachments` preceeding this call.

kurtwheeler:
Should be closed by #800!

PR #800, opened 2021-03-23
Fix validation of request documents

kurtwheeler:
## Issue Number

https://github.com/AlexsLemonade/resources-portal/issues/799

I was able to reproduce both triggering 400s by updating with valid payment info and also the request getting updated when it shouldn't have. I added fixes for both, and now the tests pass!

## Types of changes

- Bugfix


davidsmejia (approved):
LGTM!

PR #834, opened 2021-04-26
Add missing file for elasticsearch configuration

kurtwheeler:
## Issue Number

#833

## Purpose/Implementation Notes

Forgot an entire file in my PR on Friday. :facepalm: 

## Types of changes

- Bugfix (non-breaking change which fixes an issue)

PR #835, opened 2021-04-26
Fix load balancer names

kurtwheeler:
## Issue Number

#834 

Whoops, I didn't stage this change before the last PR. These names were too long as is.


davidsmejia (approved):
🔒

PR #836, opened 2021-04-27
Add duplicate resources for ones that won't cleanly cycle themselves.

kurtwheeler:
## Issue Number

#835 

The deploy failed with:
```
Error: Error deleting DB parameter group: InvalidDBParameterGroupState: One or more database instances are still members of this parameter group postgres-parameters-deployer-staging, so the group cannot be deleted
	status code: 400, request id: 1bf30bee-97ee-4311-abb3-185c68a8d9ec



Error: error deleting Target Group: ResourceInUse: Target group 'arn:aws:elasticloadbalancing:us-east-1:589864003899:targetgroup/dr-api-deployer-staging-http/ab732ef807c56e56' is currently in use by a listener or a rule
	status code: 400, request id: d5a7f717-3845-4d30-af93-d10eefc6fab1



Error: error deleting Target Group: ResourceInUse: Target group 'arn:aws:elasticloadbalancing:us-east-1:589864003899:targetgroup/dr-api-deployer-staging-https/c698a9e25dc4c417' is currently in use by a listener or a rule
	status code: 400, request id: 61411bde-fcd5-41a3-89a5-6fbc221f93c0
```

I think it's because adding the tags and changing the names requires a new resource, but AWS/terraform can't just swap them out so I have to add a new resource to use, then later delete the old resource once it's unused.

## Types of changes

### DIFF
--- a/infrastructure/database.tf
+++ b/infrastructure/database.tf
@@ -15,6 +15,22 @@ resource "aws_db_parameter_group" "postgres_parameters" {
     name = "statement_timeout"
     value = "60000" # 60000ms = 60s
   }
+}
+
+resource "aws_db_parameter_group" "postgres_parameters_new" {
+  name = "resources-portal-postgres-parameters-${var.user}-${var.stage}-new"
+  description = "Postgres Parameters ${var.user} ${var.stage}"
+  family = "postgres9.6"
+
+  parameter {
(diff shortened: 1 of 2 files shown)


davidsmejia (approved):
ooph looks good

PR #837, opened 2021-04-27
Try another way of cycling lb target groups and parameter groups.

kurtwheeler:
## Issue Number

#836 

https://github.com/hashicorp/terraform-provider-aws/issues/636#issuecomment-397459646 suggests this for the target groups. It may only work once it's already like that though so I may need yet another PR after this one.

I looked at the existing parameter groups and I think this should be right now.


davidsmejia (approved):
lgtm

PR #838, opened 2021-04-27
Tweak resource names for database parameter groups

kurtwheeler:
## Issue Number

#836 

Ok so apparently resource names are important to terraform.


davidsmejia (approved):
📛

PR #849, opened 2021-05-13
renaming problem icon svgs

davidsmejia:
## Issue Number

n/a

Some icons that were using the default coloring  (colors defined in the svg) were not updating on deploy, so this is an attempt to avoid figuring out why those files are cached so aggressively and just renaming them.

* fixes the mta icons that were named copy in the asset frop
* fixes the header logo swap from being so attention grabby
* fixes the hero svg on the landing page

## Types of changes

## Functional tests

n/a

## Checklist

## Screenshots

n/a


dvenprasad (approved):
🖌️ 🖌️ 🖌️ 

Yep looks the icons and illustrations are replaced.

Issue #851, opened 2021-05-17
Text needs to be white on hover and active


dvenprasad:
### Context

May have broken when we updated branding

### Problem or idea
The text color needs to be white in this state.

![Screen Shot 2021-05-17 at 11 42 42 AM](https://user-images.githubusercontent.com/15315514/118522027-9823be00-b709-11eb-8028-01a7e5d9ae27.png)

### Solution or next step

Tagging @davidsmejia for fixes.

PR #853, opened 2021-05-18
remove brand color issue from suggestions onmouseexit

davidsmejia:
## Issue Number

#851 

* excludes the suggestions from being grouped into the plain button rules to prevent the blue text on blue background issue
* annoyingly this means that i was unable to make the suggestions text from being the brand color.... but i dont think it's worth the added effort to rewrite the rules to make this work at this time.

## Types of changes

## Functional tests

tested on the teams section when adding a member should apply universally to all textInputs in the same way

## Checklist

## Screenshots

Screen Shot 2021-05-18 at 4 58 29 PM
Screen Shot 2021-05-18 at 5 01 38 PM

### DIFF
--- a/client/src/theme/index.js
+++ b/client/src/theme/index.js
@@ -183,7 +183,10 @@ const theme = {
           `
         ),
         applyWhen(
-          props.plain && !props.role && !props.colorValue,
+          props.plain &&
+            !props.role &&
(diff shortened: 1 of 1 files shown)


dvenprasad (approved):
screenshots look good!

🐭 🐭 🐭

Issue #810, opened 2021-03-31
Support empty search on ALL resources


dvenprasad:
### Context

Came up during beta testing

### Problem or idea

Multiple users tried to search `ALL` with an empty search expecting everything to pop up in the results. It did not work.

It was one of their first steps to gauge and discover what the portal has to offer. 
 
User behavior nugget:  the more computationally savvy users tried to use `*` to search for everything. 

### Solution or next step

It would be good if we can support this.

PR #856, opened 2021-05-19
810 support empty search for all resource types only

davidsmejia:
## Issue Number

#810 

* changes the check on weather or not to perform the search on page load to allow empty searches with offset/limit to be performed
* renames the helper that checks for clarity
* keeps the category dropdown the be persistent if the resulting search does not provide results

just to reiterate if you select a resource category/type you have to enter a term... but just clicking the search with an empty search will give you a somewhat random result set.

## Types of changes

## Functional tests

tested locally with my limited resources

## Checklist

## Screenshots

Screen Shot 2021-05-19 at 12 41 48 PM


dvenprasad (approved):
🔍 🔍 🔍  ✅

Issue #806, opened 2021-03-30
Remove Empty Organisms on Save


davidsmejia:
### Context

It is currently possible to list an empty organism.

### Problem or idea

For sanitary date we need to remove empty List/Array data types before saving.

### Solution or next step

Either add to validation or automatically remove any empty organisms on save.

PR #857, opened 2021-05-19
806 removes empty organisms on save

davidsmejia:
## Issue Number

#806 

* filters out empty organisms on save
* does not filter out for preview... it didnt seem appropriate to change data during validation

## Types of changes

## Functional tests

n/a tested listing with an extra empty organism

## Checklist

## Screenshots

n/a basically imagine everything exactly the same but on the preview there is a trailing comma and then on the resource page there is no trailing comma


dvenprasad (approved):
🦩 🦩 🦩

PR #864, opened 2021-06-01
Api Endpoints with optional trailing slashes

davidsmejia:
## Issue Number

n/a

Overwrites the trialing slash regex to match regardless of whether there is a trailing slash or not.
Note: This was because the local and staging envs were behaving differently. So while this is a good solution because we wanted optional slashes anyway... it does not explain the difference of behavior between two envs that should have been behaving the same.

Also adds in the solo organizations url to viewset register.

## Types of changes

## Functional tests

n/a tested the org update endpoint with and without a trailing slash

## Checklist

## Screenshots

n/a

### DIFF
--- a/api/resources_portal/urls.py
+++ b/api/resources_portal/urls.py
@@ -40,6 +40,8 @@ from resources_portal.views import (
 )
 
 router = ExtendedSimpleRouter()
+router.trailing_slash = "/?"
+
 router.register(r"users", UserViewSet, basename="user")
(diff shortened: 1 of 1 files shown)


kurtwheeler (approved):
This LGTM! Do you know how it interacts with the `APPEND_SLASH` setting? Should we turn that off now that we have this or does it not matter?

davidsmejia:
It shouldnt matter because append slash should only apply a redirect if there is no matching url pattern

PR #866, opened 2021-06-02
Add profiling code and profile slow orgs update endpoint

kurtwheeler:
## Issue Number

N/A trying to debug this endpoint with David.

It turned out to be timing out and then nginx couldn't find the 50x.html page so it returned a 404. Therefore this also disables the undesirable nginx error pages.

## Types of changes

## Functional tests

I ran the tests and played with the profiling stats a bit. It should be sufficient for figuring out what's wrong.

## Checklist


davidsmejia (approved):
LGTM

PR #867, opened 2021-06-02
stop saving updates before interaction

davidsmejia:
## Issue Number

n/a

* fixes an issue where after the user refresh happens it would trigger a save since it detected a change

## Types of changes

## Functional tests

n/a - loads the page without the saving event firing but toggling will save the changes

## Checklist

## Screenshots

n/a


kurtwheeler (approved):
LGTM!

PR #869, opened 2021-06-03
Stops profiling the update org endpoint

kurtwheeler:
## Issue Number

Leave the profiling code for if we need it later.

Just removes the decorator from that view.

## Types of changes

- Cleanup


davidsmejia (approved):
LGTM!

Issue #831, opened 2021-04-20
Figure out why ORCID is not redirecting correctly


dvenprasad:
### Context

Came up while trying to reproduce #830 

### Problem or idea

Signed up on ORID via the invite link. Once I create my ORCID and authorized access to ALSF, it redirected me to the landing page instead of letting me continue with the invite flow.

PR #870, opened 2021-06-03
831 fix redirect after login

davidsmejia:
## Issue Number

#831 

* adds a flag ref in the `ResourcesPortalContext` that determines if /account should do the temp redirect

The `account/index` page has a temporary redirect to `account/basic-information`. This was firing after the redirect to the place the use left off before signing in which was overriding that functionality. This PR adds a flag to ignore that temp redirect if they just logged in. It's not great but it works and will be removed when that page no longer has the redirect on it.

## Types of changes

## Functional tests

n/a 
* tested logging in from search and went back there
* tested the /account link and was redirected to basic-information
* tested with create-account flow

## Checklist

## Screenshots

n/a


kurtwheeler (approved):
LGTM!

Issue #871, opened 2021-06-03
Search results repeat on different pages.


dvenprasad:
### Context

I did a blank search and got 4 pages worth of results. (which is also weird because it said it was showing 10 of 30 results) 

### Problem or idea

When I navigated to page 2, some results which were on page 1 was also present. As I went to the next page, there were again, some repetitions from the previous page. I did not even see some of the resources which I expected to see in the results. 

I'm not sure if this error happen with searches with a search term (cause there aren't enough resources for me reproduce that case)

PR #875, opened 2021-06-04
871 search pagination is offset based not page based

davidsmejia:
## Issue Number

#871 

* adds helpers to convert pages to offsets and back
* implements them in the `SearchResultOffset` component
* goes to top of window when you change pages
* updates the component that shows how many per page to say that so it is less confusing

## Types of changes

## Functional tests

tested with various offsets and limits

## Checklist

## Screenshots

n/a but here

Screen Shot 2021-06-04 at 2 03 56 PM


kurtwheeler (approved):
Nice! LGTM!

PR #888, opened 2021-06-16
fixes issue with unable to edit resource and flash of error after save

davidsmejia:
## Issue Number

n/a

Setting the resource to `undefined` was causing an error to flash before navigating away from the page.
Setting it to `{}` fixed the flashing error but prevented the local storage value from being overwritten using the initialValue argument in `useLocalStorage`.

This change cleans up the storage without the error before navigating away from the page.

## Types of changes

## Functional tests

n/a tested editing a resource, then navigating away and editing another

## Checklist

## Screenshots

n/a


kurtwheeler (approved):
Makes sense to me!

PR #891, opened 2021-07-06
Update the AWS key pair used for SSH onto instances.

kurtwheeler:
## Issue Number

N/A doing some key rotation


davidsmejia (approved):
🔒

PR #913, opened 2021-08-16
Make the API depend on its SSH key so it actually rotates when the key does.

kurtwheeler:
## Issue Number

#912


davidsmejia (approved):
🔥

PR #924, opened 2021-11-24
Taint the API instance so we can test SSL cert persistence.

kurtwheeler:
## Issue Number

#923 

We stored the SSL cert in S3, but when I redeployed to see if it would use it successfully the API instance didn't get cycled, so I couldn't tell.

I'll make another PR to remove this once it's tested.

## Types of changes

- Temporary test


davidsmejia (approved):
LGTM

PR #926, opened 2021-11-24
Update which token is used by the docs workflow.

kurtwheeler:
## Issue Number

I think I missed this in https://github.com/AlexsLemonade/resources-portal/pull/921

This should have changed. I also think the one used for linting was unnecessary.

## Types of changes


davidsmejia (approved):
LGTM

Issue #954, opened 2022-09-02
Terraform Builds Not Working


davidsmejia:
### Context

There seems to be an issue when fetching the AMI for the API instance. Nothing seems to be returned by the query anymore. It is possible this image has been deprecated.

### Problem or idea

Fix the deploy.

### Solution or next step

Either manually try to query the AMI via the aws cli, or test with dry-run terraform apply

davidsmejia:
Closed with #955 and #956

PR #956, opened 2022-09-07
pull docker from their repos

davidsmejia:
## Issue Number

#954 

- installs docker-ce and docker-ce-cli from docker instead of ubuntu repository

## Types of changes

## Functional tests

n/a

## Checklist

## Screenshots

n/a

### DIFF
--- a/infrastructure/api-configuration/api-server-instance-user-data.tpl.sh
+++ b/infrastructure/api-configuration/api-server-instance-user-data.tpl.sh
@@ -22,7 +22,7 @@ cp nginx.conf /etc/nginx/nginx.conf
 service nginx restart
 
 # install and run docker
-sudo apt install apt-transport-https ca-certificates curl software-properties-common -y
+apt-get install apt-transport-https ca-certificates curl software-properties-common -y
 curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
 add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu jammy stable" -y
 apt-get install docker-ce docker-ce-cli -y
--- a/infrastructure/api-configuration/api-server-instance-user-data.tpl.sh
+++ b/infrastructure/api-configuration/api-server-instance-user-data.tpl.sh
@@ -24,7 +24,7 @@ service nginx restart
 # install and run docker
 apt-get install apt-transport-https ca-certificates curl software-properties-common -y
 curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
-add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu jammy stable" -y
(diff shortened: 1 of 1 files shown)


infrastructure/api-configuration/api-server-instance-user-data.tpl.sh
```diff
@@ -22,8 +22,10 @@ cp nginx.conf /etc/nginx/nginx.conf
 service nginx restart
 
 # install and run docker
-apt-get install docker-ce -y
-service docker restart
+apt-get install apt-transport-https ca-certificates curl software-properties-common -y
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
+add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu jammy stable" -y
```
arkid15r, line 27:
```suggestion
add-apt-repository "deb [arch=$(uname -m)] https://download.docker.com/linux/ubuntu jammy stable" -y
```

infrastructure/api-configuration/api-server-instance-user-data.tpl.sh
```diff
@@ -22,8 +22,10 @@ cp nginx.conf /etc/nginx/nginx.conf
 service nginx restart
 
 # install and run docker
-apt-get install docker-ce -y
-service docker restart
```
arkid15r, line 26: I assume it does start automatically after the installation process is over?

arkid15r (approved):

nozomione:
Just a thought, in the [Install Docker Engine on Ubuntu](https://docs.docker.com/engine/install/ubuntu) doc; 
- before installing the docker engine, it runs [`apt-get update`](https://www.cyberciti.biz/faq/what-does-sudo-apt-get-update-command-do-on-ubuntu-debian)(although not necessary).
- for adding GPG key, since [`apt-key` is deprecated](https://manpages.ubuntu.com/manpages/jammy/man8/apt-key.8.html)("*apt-key(8) will last be available in Debian 11 and Ubuntu 22.04.*" and writes a warning message to stdout), instead it uses the [gnupg](https://installati.one/ubuntu/22.04/gnupg/)'s [`gpg`](https://www.gnupg.org/documentation/manuals/gnupg/Operational-GPG-Commands.html) command. For this, perhaps it might be a good idea to use the gnupg?

(Additional resource) 
 [How To Handle apt-key and add-apt-repository Deprecation Using gpg to Add External Repositories on Ubuntu 22.04 - DigitalOcean](https://www.digitalocean.com/community/tutorials/how-to-handle-apt-key-and-add-apt-repository-deprecation-using-gpg-to-add-external-repositories-on-ubuntu-22-04)
 
 Thank you, David!

infrastructure/api-configuration/api-server-instance-user-data.tpl.sh
davidsmejia: yeah it usually does this is just in case there is a reconcilable error

